PR #5, opened 2015-09-15
Update rpi.md - optional watchdog

danamlewis:
Just marks watchdog as optional.

See if @bewest can fix my formatting on the numbers? (example, watchdog and everything that should be 4-9 became 1s)


danamlewis:
Closing since @channemann pulled into #9 . (Also thanks for the formatting info, think we're going with Chris's recommendations of headers to stay away from numbered lists for the rest of the docs.)

PR #7, opened 2015-09-16
Update README.md

danamlewis:
More README additions, (similar to #31 on openaps-js). @channemann @scottleibrand etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,37 @@
 
 Welcome to the documentation for [openaps](https://github.com/openaps/openaps).
 
+This is part of a series of tools to support a self-driven DIY
+implementation based on the OpenAPS reference design. The tools may be
+categorized as *monitor* (collecting data about environment, and
+operational status of devices and/or aggregating as much data as is
+relevant into one place), *predict* (make predictions about what should
+happen next), or *control* (enacting changes, and feeding more data back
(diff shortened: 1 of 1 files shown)


bewest:
What do you think about the meta-colophon bit about this book being written in markdown.  Seems like we can remove that or move it to colophon section?

bewest:
This might serve as nice basis for exploring common snippets to include in all the project READMEs, nice.

danamlewis:
Hmm, maybe remove it? It looks a bit weird in github, but could go either way.

channemann:
The initial addition of that weird "written in markdown" section was because the same page was going to show up as the readme in GitHub and the intro page in the GitBook version, and I wasn't quite clear on what the link would end up being (it was my first PR with a GitBook). I'd go ahead and remove it.

If this is going to go into all of the openaps- repo readme files, is it worth cross-linking them (i.e. having an index of sorts)? Then you could label them each as _monitor_, _predict_, and _control_ in that index.

I would also look to put a little more stress on the cover-our-ass portion. Not sure what the actual language should be (isn't Ed a lawyer?), but some emphasis on that line might be appropriate.

danamlewis:
Closing since @channemann's #9 pulls my additions in to his big PR there.

PR #15, opened 2015-10-04
Small edits

tghoward:
A few minor typos in the Overview section and intro. Perhaps the most significant is the 404 error to the link to the commercial efforts subpage (currently written as its own page). Looks like you *can* link down to the sub heading (tried it manually by typing into url) but I'm not sure how. So I just changed the link to the entire page. 

hope this helps. 
Tim


danamlewis:
Excellent, @tghoward - thanks for the review on these. I think the 404 is because I had moved the commercial page to be within the other projects page...not sure if your fix will work, but I'll merge and then re-fix if needed if it doesn't fully resolve that one.

danamlewis:
Fixed in another patch, but for future reference, to refer with a link within the files, should be referenced as ../<foldername>/<filename>#<subsection__or_headername> to create the link. In this case it was ../Other-projects/other-projects.md#commercial-aps-efforts

PR #17, opened 2015-10-13
Update rpi.md

scottleibrand:
Get rid of old references to numbered subsections, and clean up language a bit.


tghoward:
Looks good to me! I'm working on some more updates to this page, but they don't conflict with what you've got here.

PR #22, opened 2015-10-25
update setuptools and install npm; do git setup after python-git

scottleibrand:


bewest:
What about something like this to make installing all the packages easier.
There's no config or anything in these steps, so might as well reduce it to one.

## Install openaps and dependencies

### Using the package manager

This is the recommended way to install:

`curl -s https://gist.githubusercontent.com/bewest/b6c41e05148117802f51/raw/fa0e8dd95b477f54695d9d80d4f59cb282808096/quick-packages.sh | sudo bash -`

This downloads the content from [this gist](https://gist.github.com/bewest/b6c41e05148117802f51) and uses it to install all the dependencies in one step.

### Installing from source

It's possible to use the package manager to install development branches.  If you are hacking on the code, you'll need a way to develop using versions you control.  Here's a quick way to do that:

`curl -s https://gist.githubusercontent.com/bewest/b6c41e05148117802f51/raw/fa0e8dd95b477f54695d9d80d4f59cb282808096/quick-src.sh | sudo bash -`

scottleibrand:
If we want a one-step install of openaps, oref0, and all dependencies, I wouldn't mind using an install script like that, but we'll need to make it part of one of the official repos and keep in maintained, rather than relying on a gist.

We should probably make the install script use && to chain all the commands together that depend on each other, so when my apt-get download stalls and I ^C to restart it, it doesn't try to install things out of order.

If we go down this route, I could also do the same to automate the wifi and watchdog optional stuff in https://openaps.gitbooks.io/building-an-open-artificial-pancreas-system/content/docs/Setup/rpi.html

bewest:
Wow, nice.

Don't forget to `npm install -g json` also.

PR #23, opened 2015-10-27
Python setuptools

scottleibrand:
@eszcloud is going to test this and make sure it works properly before we merge it.


eszcloud:
Worked for me so far with a fresh install. I didn't need to run the install line twice.

PR #31, opened 2015-11-16
Help links for unzipping .img, backing up with github

jmatheson:


scottleibrand:
Merged this into dev instead of master via #32.

PR #32, opened 2015-11-16
Re-target @jmatheson's #31 at dev

scottleibrand:


scottleibrand:
This merges #31 into dev instead of master.

PR #33, opened 2015-11-16
Update rpi.md

logichammer:
Small change as to where Path 1 and 2 are referenced


logichammer:
Your changes make more sense / better organized now...

PR #35, opened 2015-11-16
Pulling re-org changes from dev to master

danamlewis:
Re-org of docs overall to reflect "rainbow' outline to better match oref0 as it is today:
*Adds phase visual image to intro page
*Moves phase0-4 to intro section, re-word and updated to reflect image outline.
*Created some new sub pages to reflect steps in phases 0-4
*Added a lot that needs to be edited/reviewed/added to in phase 3 and 4 (!)
*As always, asking people to review and edit how they want to be listed in the people section, if at all
*Added a FAQ section that I intend to add to more later, but is a good home base for usability-type info
-I think I got the links that needed to be updated as a result of all the file and folder re-naming, but as always would be good for someone to do a deep check on those.
-More PRs and fixes welcome to build on these re-org changes :)

PR #36, opened 2015-11-17
Rainbow connection

channemann:
* Removed some redundancy in the phase organization by consolidating the sections in the introduction into the first page of each phase section.
* Fixed issues with images
* General rewording, typo fixing, etc
* Added notes on AAA batteries and cases to hardware section
* Added a troubleshooting section with common linux commands (still needs work)
* _started_ some reformatting of Phase 2, but stopped short (requires further discussion)
* Moved document conventions to separate subsection
* Split out ToC
* Other modifications and fixes I can't quite remember

Not sure what the automatic merge issue is; let me know if I can help alleviate that


scottleibrand:
Looks pretty good to me. Noticed a couple really minor things I can fix when resolving the merge conflict. Only regression I see is that you got rid of the TOC embedded in the README, so now there's no good way to link to all the content without using the gitbook. We should probably readd that, and add a bunch more links within the new phase sections to each of their component pages. It should be possible for someone who just lands on the repo to read the first page, start clicking to follow links and read things in order, and not miss any content.

channemann:
Yeah, I went back and forth on the TOC for that same reason—having it on the landing page. The intro had just become so long and I thought having an easily-found TOC (i.e. one that had an explicit sidebar callout) would be good.

I get what you're talking about with the pages being linked to each other, but is there any particular reason for someone not to view it in the gitbook?

scottleibrand:
Gitbook doesn't work at all for reviewing things in a dev branch, for one thing. We have had discussions about ditching gitbook entirely due to unexplained update delays etc.  And some people just won't notice the gitbook link. All of those point to making the repo self-contained, with gitbooks an optional nice UI. 

-Scott




On Tue, Nov 17, 2015 at 9:26 AM -0800, "Chris Hannemann" <notifications@github.com> wrote:












Yeah, I went back and forth on the TOC for that same reason—having it on the landing page. The intro had just become so long and I thought having an easily-found TOC (i.e. one that had an explicit sidebar callout) would be good.



I get what you're talking about with the pages being linked to each other, but is there any particular reason for someone not to view it in the gitbook?

danamlewis:
Finally had a chance to review (all) of this great work! I like it. A few thoughts:

-what about removing or instead embedding the first image of the setup on the Intro page? that takes a lot of space, and it's included elsewhere. I think that would give us more space, and I'd like the TOC on the home page at the bottom since more people are coming to github rather than using gitbooks to nav through this, and then beyond phase 0-4 they can see all the great stuff as a part of these docs.
-Separate TOC is also good to have.

With that, I'll give this another +1. (We'll go ahead and merge #35 and then can talk about how to proceed with these changes so they're compatible).

danamlewis:
Closing because #38

PR #37, opened 2015-11-19
list of useful apps

eszcloud:
list of a few apps that are useful for interacting with RPi from smartphone


danamlewis:
Great, thanks! We have some merge conflicts that @scottleibrand will resolve (based on my re-work in #35) and then can incorporate (otherwise looks good, +1)

PR #40, opened 2015-12-01
This is the fast stuff I can add

logichammer:
Marked a few questions with DL so you could find them quickly.

I have more I need to do with my setup (add in push notifications / understand how and why it has certain recommendations so I will have more to add later).


scottleibrand:
there is a read_current_basal_profile or some such that reads whichever pattern is active. That's the one we want.

clock.json is used as the "now" for IOB calculations.

logichammer:
Perfect...will update docs (and my own install!) with this!

logichammer:
Updated and ready to merge into master if you like what you see.

danamlewis:
Looks good, thanks @logichammer

PR #4, opened 2015-09-10
Updating raspberry pi setup instructions

danamlewis:
Includes:
1. Minor edits and revisions of links, language, etc.
2. Added section (3b) how to do multiple wifi setup (for hotspot and home, suggested) from a Mac
3. Added note about console cable and good how-to link for using that with Mac
4. Edited watchdog section to run now, w/ @channemann's notes from his gist on that step

PR #45, opened 2015-12-06
{WIP} Preparing to loop

oskarpearson:
Initial work-in-progress notes on preparing to loop.

Comments / assistance with this section is welcome

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -15,10 +15,11 @@
 * [Phase 2: Build a Manual System](docs/Build-manual-system/considerations.md)
    * [Using oref0 Tools](docs/Build-manual-system/Using-oref0-tools.md)
    * [Creating a Loop and Retry Logic](docs/Build-manual-system/loop-and-retry-logic.md)
-* [Phase 3: Automate Your System](docs/Automate-system/considerations.md)
+* [Phase 3: Preparing to Loop](docs/Preparing-to-loop/overview.md)
+* [Phase 4: Automate Your System](docs/Automate-system/considerations.md)
    * [Creating a Schedule](docs/Automate-system/create-schedule.md)
    * [Validating and Testing](docs/Automate-system/validate-output.md)
-* [Phase 4: Iterate and Improve](docs/Iterate-improve/improvement-projects.md)
+* [Phase 5: Iterate and Improve](docs/Iterate-improve/improvement-projects.md)
 * [Resources](docs/Resources/resources.md)
    * [Technical Resources](docs/Resources/technical-resources.md)
    * [Troubleshooting](docs/Resources/troubleshooting.md)
@@ -26,4 +27,3 @@
    * [Other Projects, People & Tools](docs/Resources/other-projects.md)
    * [FAQs](docs/Resources/faq.md)
    * [Glossary](docs/Resources/glossary.md)
-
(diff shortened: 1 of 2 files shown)


oskarpearson:
Capturing some comments from Gitter chat:

>Hi Oskar,
>Nice job on adding the preparing the loop section. I only have a few other thoughts on your additions:
>1) One question I think people will have is how much to follow what the openaps system recommends vs to continue what they normally do. So if at this stage, they are manually running the system and looking at the recommendations. The question I always had was – okay do I follow this thing exactly and see what happens or do I manually intervene and at what point should I do so. In particular, I think for lows we should tell people to not depend on the system at all for recoveries from low. Per your point #4, the system cannot do very much to resolve a low. I would tell them to treat lows as normal and see how it affects the system.
>2) When to do it - I have found the best time to test is when there is little to no insulin on board or carbs on board and when the person is going to be close to the openaps rig. For a parent testing the system on a child, this probably means late at night. For an adult, this may mean first thing in the morning. It was much easier for me to understand the basics to the system, later at night once most of the insulin had cleared out of my son’s system . In addition, it was one of the few times that he would stay put in one place enough to test. During the day time when he has 5 or 10 + units on board from always eating (a teenager), there were few times that the openaps system would even give a recommendation.
>3) Setting your targets - People are probably going to want to understand a bit more what you mean by “high and wide” in section about setting your targets. You do a great job in the basal section talking about in the beginning setting max basal at 1.3x the normal basal rate. Something similar would be useful. For example, normally I have both the high and low set to 120 for my son. I switched it to a high of 140 and low of 120 for some of the tests. But do you think a high of 180 and low of 150 is better? I know it will be different for each person but to give some examples would be helpful for the new folks.
>4) Understanding Transition Range Limits - I might give people rough estimates on the effective distance of the carelink stick. I found that the max distance with nothing blocking is 9 feet. But with clothes / blankets / body laying on the pump, the distance can drop to 1 foot or less. A picture or diagram of how people have run it under the bed would be helpful. This was an early major point of frustration for me as I eventually had to run it on a battery and put it in a bag under the covers with him to get it to work at night.
>5) I other thing I found useful was to set up a script that runs everything but doesn't enact anything and to send the results from the loop to a file they can review. This helped me a lot to be able to go back and see for each time period what was happening and what failed and what the inputs were. I used something like this:
date >> record.txt (so that I can see what time each loop ran)
openaps run-loop-not-enact 2>&1 | tee -a record.txt
I would then look at the record.txt so I could see what each pass did. Tools like what @jasoncalabrese is working on to put this into Nightscout is very effective as well
```

danamlewis:
@oskarpearson let me know what you think re: phase 3 inclusion? Two paths: if you like that idea, you can resubmit with that structure. Or if you want, I can create a branch, PR this into there, and make those structure changes (put to phase 3) and then add some of my edits, including some of the specifics from @eyim feedback and then PR it to dev so you can add more there. Let me know?

oskarpearson:
@danamlewis I'm happy with it wherever you'd like it :)

You're welcome to make changes while this is in-flight. I'm only likely to get to this again later tonight, or possibly tomorrow. I'll make the above changes if you don't get there by then.

Thanks!

Oskar

danamlewis:
Lots of merge conflicts in this now, exacerbated by the fact that we just re-worked dev :). Solution: I am going to close this, and pull the content over into the PR I am making into the updated dev branch. High level, some of this content will update the Phase 3 overview page, based on earlier discussion. Some of the other content I will put into a new page under phase 3 about questions by new loopers.

PR #47, opened 2015-12-14
add comment - install package manager first

eyim:
need to run package manager before install the dev branch stuff

My first pull request so doing something easy.

### DIFF
--- a/docs/Automate-system/validate-output.md
+++ b/docs/Automate-system/validate-output.md
@@ -23,3 +23,76 @@ If you'd like to help out with defining all the desired behaviors in the form of
 ###How to add more test cases:
  
 We'll want to cover every section of the code, so if you see a "rT.reason" in bin/oref0-determine-basal.js that doesn't have a corresponding "output.reason.should.match" line in an appropriate test in tests/determine-basal.test.js, then you should figure out what glucose, temp basal, IOB, and profile inputs would get you into that section of the code (preferably representing what you're likely to see in a real-world situation), and create a test case to capture those inputs and the desired outputs.  Then run the tests and see if your test passes, and the output looks reasonable.  If not, then modify your test case accordingly, or if you think you've found a bug in determine-basal.js, ask on Gitter.
+
+
+### Nightscout Integration
+
+Integrating Openaps with Nightscout is a very helpul way to understand what Openaps is doing in a visual way and can be accessed through a web browser rather than logging into the Raspberry Pi and looking at the logs.  
+
+There are a few changes to your Nightscout Implementation:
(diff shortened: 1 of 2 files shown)


eyim:
When I run the Installing from source script by itself, it fails.  However when I run it after the package manager it seems to work fine so adding that to the docs.  I am not sure if that the expected behavior.

scottleibrand:
The two scripts install identical dependencies, so you shouldn't have to run both. But I think one of them might be out of order or something, because you have to run it twice. You can either run the two scripts, or retry the same script a second time. We need to figure out what exactly is failing, why, and how to fix the script to work the first time.

eyim:
Scott - how would you me to report bugs / errors like this.  I had already mentioned it on gitter.  Is there a place to list them so that the group can decide which and when to fix them?  My biggest concern is that right now if someone is following the openaps docs they will be getting a bunch of errors with no explanation or idea how to get around it.

scottleibrand:
I think we need to do two things.  For this PR, we should change the verbiage to mention that you need to run the command you want twice if it doesn't succeed the first time.  Then we also need to open an issue (https://github.com/openaps/docs/issues) to track investigating and resolving the problem in the script that is causing us to have to do that.  If you have the time and patience, you could then start working that issue by doing a fresh image and walking through the script one command at a time until you can see exactly what's failing.

eyim:
Great.  I have submitted the issue ticket and put in the error message I get when trying to run the script.

I don't understand your comment about changed the verbiage to run the command twice if it doesn't succeed the first time.  Basically, the only way I have found to run the install from source script is to first run the Using the package manager script.

As you saw today, the Using the package manager script does not install the current openaps version.  So right now there is no simple way from the documentation to get a working version of openaps without first running the package manager script and then running the installing from source script on after it.

Obviously the preference is to get both scripts fixed but until then it may make sense to just have people run both scripts together to get an working openaps system.

scottleibrand:
Have you tried running quick-src.sh twice in a row, rather than running quick-packages.sh followed by quick-src.sh? The issue I saw earlier was resolved by doing that, but not sure if it's the same one.

eyim:
Yes I tried to run quick-src.sh twice in a row but it didn't help.  Still failed.

eyim:
Here is the information on integrating Openaps and Nightscout.  I wasn't sure the best place to place this so I put it in validating and test.  Please review and make sure it makes sense

eyim:
Hmm - hope this pull request went through

scottleibrand:
This is good stuff.  I'm going to merge it now, so we have something to point people to, and then we can work on further revisions (for example, providing people exact commands to run so they don't have to edit ini files) after that.

PR #62, opened 2016-01-02
Nightscout Integration Edits

audiefile:
The original did not work exactly for me, particularly editing openaps.ini directly to add the ns-upload device. This is what I believe the command should be (I didn't take notes). Also includes other minor edits (typos, etc).


scottleibrand:
Applied these changes manually, since they no longer merge.

PR #69, opened 2016-01-04
Dev to master

scottleibrand:
Split out NS visualization from validation and testing.  Incorporate changes from @audiefile on top of @eyim updates in master.  Incorporate @oskarpearson's #45.


danamlewis:
Approving based on prior review. Still need some attention on the Visualization section for people to validate all the commands work for a from-scratch setup, and to better match the overall style of the rest of the docs, but fine generally content-wise as a starting place.

PR #70, opened 2016-01-05
Add xdrip info

audiefile:
Added xdrip as an option for cgm data.


danamlewis:
Closed in favor of #72 targeting to dev instead.

PR #72, opened 2016-01-05
Add xdrip info

danamlewis:
Added xdrip as an option for cgm data. (Retargeting to dev)


danamlewis:
Now merging, will add a follow up PR with some formatting tweaks, but good content to help people get started. Thanks @audiefile !

PR #77, opened 2016-01-20
Dev

dm61:
I've made some updates to the visualization section based on my recent experience. I think someone else should take a look at it, especially the "Configuring and Uploading OpenAPS Status" subsection - I have not been able to figure out how to add ns-upload as an openaps device that would produce the desired entries in the openaps.ini file. Included is my best attempt. There are other details I am not so sure about - the treatment upload subsection is largely based on copy and paste from Scott's script and I do not think I fully understand all details.

### DIFF
--- a/docs/Automate-system/vizualization.md
+++ b/docs/Automate-system/vizualization.md
@@ -60,7 +60,7 @@ args = <br>
 Then use the `ns-status` device to add the `monitor/upload-status.json` report, which you may do as follows:
 
 ```
-$ openaps report add monitor/upload-status.json JSON shell ns-status monitor/clock-zoned.json monitor/iob.json enact/suggested.json enact/enacted.json monitor/battery.json monitor/reservoir.json monitor/status.json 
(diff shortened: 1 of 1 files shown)


TC2013:
Line 63 should read: 
openaps report add monitor/upload-status.json JSON ns-status shell monitor/clock-zoned.json monitor/iob.json enact/suggested.json enact/enacted.json monitor/battery.json monitor/reservoir.json monitor/status.json

ns-status and shell are turned around

jasoncalabrese:
Would be great to add some info about the new pump plugin here, it's in the NS dev [readme](https://github.com/nightscout/cgm-remote-monitor/blob/dev/README.md#built-inexample-plugins) now

screen shot 2016-01-27 at 8 33 25 am

dm61:
I've added a brief description of the new pump plugin, with an example of how to add reservoir and battery, and with a link to the NS readme. The pump plugin is very useful - thanks!

scottleibrand:
:+1: I think this is better than what we have now, and ready to merge.  @jasoncalabrese can you PR your suggested changes to dev after that?

danamlewis:
@dm61 thanks for the further additions and great work on this. Think we'll go ahead and merge to dev, and then try to figure out what is still tripping people up, but this is way better than current dev regardless :)

PR #80, opened 2016-01-28
Dev

logichammer:


logichammer:
This is a comment

logichammer:
FYI, let me know if you want me to size the images better so they aren't so huge...Markdown looks to just thrown whatever the default image size in but it looks like I can add real img tags and size things that way if preferable.

scottleibrand:
With the BGI caveat, I think this now looks good enough for a merge to dev.

danamlewis:
Hey @logichammer - content is all good! I have some questions about where content is placed. 

- I like your annotated image of the hardware setup, and we've previously had one embedded, but got feedback that it prevents people from getting to the critical written content in the readme. How about revising the second sentence of the second paragraph (currently "(Click here to view an image of the common physical components of an OpenAPS setup.)") to something like "(Here are two visuals to show you what the physical hardware components of an OpenAPS setup look like - version A is without labels; version B contains labels to describe the parts.)" (Yours obviously would be Version B, or however you want to differentiate between the two, and we'd hyperlink them instead of embedding.)

- Your NS explanation is great! Did you see the added Visualization section, though, in phase 3? Wondering if you could review it (https://github.com/openaps/docs/blob/dev/docs/Automate-system/vizualization.md) and see what you think about integrating all of your content in that page? (I'd vote to do that for now, and we can discuss if the Visualizations/Monitoring page should get moved from under Phase 3 to Phase 2, but would probably be separate from using-oref0-tools regardless, since that's a huge page in of itself.)

Thoughts?

logichammer:
FYI, took Ella snow camping Friday afternoon and just now getting back into
things...I should be able to make these changes tonight!

Personally, I think the documentation needs some images out front to
inspire folks / break up some of the dense technical jargon...show them the
reward for figuring out everything...heck, a NS with Basal graph showing
perfect numbers after doing something like eating pizza would be super
inspiring.  If scrolling is a barrier for folks, this project might be too
much. You could say the same about making folks having to click through to
see images but I think showing folks what the reward is up front, will
inspire them to delve deeper.  Seeing one of Dana's graphs was the reason I
got all hot and bothered to sit even more in front my computer.  "What if
those could be Ella's number?".  Just my two cents.  I'll make them link
but wanted to give you my take on it.

No,  totally missed the piece about Visualization (I want to add the phrase
Nightscout in there because that's what I was scanning for).  I'll see
about merging everything.

Thanks folks and see you soon!

Matt

PS In person questions I'll have:
Can you tell us more about the two other routes (increased range and
miniaturization)?
Are you OK with promotion / getting the word out?  If so, guidelines?
Guidelines for onboarding new folks that I personally know with T1D?  I'm
following whatever rules you guys layout...not going to go rogue.
Other areas that need documentation?  The two different routes referenced
above?

danamlewis:
@logichammer - this is now super out of date with where docs are. I think we've got most of this folded in, but would love for you to go look at the next iteration of dev (coming shortly, as a PR from dana-dev) to see if any of the NS stuff you worked on here should be added into the NS visualization section (now much higher in the phase flow). That'll be the place to open a new PR to and adjust from there. Thanks!

PR #82, opened 2016-02-01
Mm cgm additions1

tghoward:
Just completed a full re-install and refactoring. This PR provides instructions for a way to handle MM CGM data. I also have a tweak to `determine-basal.js` that would make the second section (json call) unnecessary but would break `determine-basal.js` for dexcom data (sure there's a fix for that).   I originally built the instructions for using loudnate's glucosetool, but the most recent changes to it (glucosetool) broke it in other ways for me.   I'll understand if you don't want to accept all of this PR - seems kind of funky to me.


tghoward:
dang, meant to do this to dev.

PR #83, opened 2016-02-01
Mm cgm additions1

tghoward:
Targeting dev this time. Need to talk about the solution provided here for filtering. Filtering is something loudnate's glucosetools does, but the new changes to glucose tools messes up the dates for other parts of Oref0.


tghoward:
Reverted back to describing how to use glucosetools. Much happier with how it is described now.

danamlewis:
Looks good @tghoward ! Thx for folding this in.

PR #90, opened 2016-02-20
Update using.md

joannestevens:
Please can you add as outlined above ... it wasn't immediately clear to me that's what I needed to do


danamlewis:
This was noted in the docs by someone else, so resolving this PR - thanks!

PR #91, opened 2016-02-27
Wifi troubleshooting tips

jmatheson:
New page to for wifi tips and tricks.


docs/Resources/wifi.md
```diff
@@ -0,0 +1,25 @@
...
+You can add a line to your cron that will check to see if <YOURWIFINAME> is avaiable and automatically switch to it if you are on a different network.
+  * Log in as root ```su root```
+  * Edit your root cron ```crontab -e```
+  * Add the following line ```*/2 * * * * ( (wpa_cli status | grep <YOURWIFINAME> > /dev/null && echo already on <YOURWIFINAME>) || (wpa_cli scan > /dev/null && wpa_cli scan_results | egrep <YOURWIFINAME> > /dev/null && udo pa_cli select_network $(wpa_clilist_networks | grep jsqrd | cut -f 1) && echo switched to <YOURWIFINAME> && sleep 15 && (for i in $(wpa_cli list_networks | grep DISABLED | cut -f 1); do wpa_cli enable_network $i > /dev/null; done) && echo and re-enabled other networks) ) 2>&1 | logger -t wifi-select```
+
+#I am having trouble consistently connecting to my wifi hotspot when I leave the house (iPhone)
+1. When you turn on your hotspot it will only broadcast for 90 seconds and then stop (even if it is flipped on). So, when you leave your house you need to go into the setting and flip it on (or off and on again if already on). Then you need to leave this screen open until you see your rig has connected.
```
scottleibrand: Don't need to flip it off. Just leave the screen open until it connects.

docs/Resources/wifi.md
```diff
@@ -0,0 +1,25 @@
...
+2. If your phone knows about wifi networks that your rig does not, your phone may connect to that wifi and in turn drop the hotspot. Make sure your phone does not connect to other wifi networks (coffee shops, a friends house, etc). You man need to forget the networks. This can also be tricky if your computer and iPhone share a network list.
```
scottleibrand: If you have something connected to an iPhone hotspot, it won't disconnect it to connect to wifi. Can probably just delete this one, as I think it's covered by 1.

docs/Resources/wifi.md
jmatheson: Good to know

docs/Resources/wifi.md
jmatheson: What specifically are you commenting on? 
If my phone knows about a wifi network, but my Edison does not. And my phone gets into range...it will connect to the wifi. phone can not send and receive wifi at the same time, so it drops the Edison.

scottleibrand: What phone is that? I specifically do *not* see that behavior with iOS 9. As long as the blue Hotspot bar is present, it won't connect to wifi unless I manually force it to.

docs/Resources/wifi.md
jmatheson: Verified that this works. will edit.

docs/Resources/wifi.md
jmatheson: Will test this and adjust language when verified

jmatheson: Verified you are right and removed it

jmatheson:
I think this PR is ready to be merged now.

PR #93, opened 2016-03-04
Update Using-oref0-tools.md

live4sw:
Simplified so that there is a 120 second sleep whenever preflight fails.


scottleibrand:
Applied this to via https://github.com/openaps/docs/commit/736b744c164c631964065ecda88e556b27435848

PR #100, opened 2016-03-22
Wip/rst

bewest:
Use sphinx/readthedocs to build the docs.

PR #102, opened 2016-03-24
Dev

bewest:
Merging into master for a "release".


bewest:
Doing after https://github.com/openaps/docs/pull/100 to include sphinx, part of transitioning away from gitbook.

bewest:
Looking good: http://openaps.readthedocs.org/en/dev/

PR #106, opened 2016-03-24
symlink docs again (in reverse this time)

bewest:
This provides a `./docs/` symlink at the root into `sources/docs`, pretty much like it was. :blush: 

Also, this demonstrates using personal fork to stage/prep PRs going forward.
http://bewest-openaps-docs.readthedocs.org/en/bewest-dev/


bewest:
done with this PR.

PR #137, opened 2016-04-30
update to hardware.md

billstackpole:
Added Zendure 2.1a 10000 mAh inline charging battery pack.


danamlewis:
Pulled it in via #145.

PR #142, opened 2016-05-03
Added Clarification about unit conversion

sjolundjohn:
Added clarification about how unit conversion process works and also split the creation of `profile`report into different steps dependant upon if unit is used or not.


scottleibrand:
Retargeted to dev with #143 and merged. Closing this one.

PR #147, opened 2016-05-04
Update hardware.md

xpucuto:
My pump is 754 (2.6A) , so i am adding it to the list.


bewest:
Moved to dev over here: https://github.com/openaps/docs/pull/148

PR #150, opened 2016-05-06
Update rpi.md

xpucuto:
As previously discussed with Scott if it stays NO you can lock out .


scottleibrand:
See my line note in #151

PR #151, opened 2016-05-06
Update rpi.md

xpucuto:
Ignore the first pull request, this one is with more details for non-programers like me.

### DIFF
--- a/docs/docs/walkthrough/phase-0/rpi.md
+++ b/docs/docs/walkthrough/phase-0/rpi.md
@@ -229,15 +229,16 @@ Now you should be able to log in without a password. Try to SSH into the RPi2 ag
 ### Mac and Linux
 If you don't already have an ssh key, then on your local computer (*not* on the Pi), run `ssh-keygen` (keep hitting enter to accept all the defaults).
 
-If you created a new key identity and accepted all of the defaults, then the name of the newly generated identity will be `id_rsa`. However, if you set a custom name for the new identity (e.g. `id_mypi`), then you will need to add it to your local ssh keyring, via `ssh-add ~/.ssh/id_mypi`.
+If you created a new key identity and accepted all of the defaults, then the name of the newly generated identity will be `id_rsa`.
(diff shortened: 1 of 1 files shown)


docs/docs/walkthrough/phase-0/rpi.md
```diff
@@ -257,7 +258,7 @@ to
 
 ```
 PermitRootLogin no
-PasswordAuthentication no
+PasswordAuthentication yes
```
scottleibrand: This change doesn't make sense. The whole point of this (optional) section is disabling PasswordAuthentication if you wish to do so. If people are blindly following directions without considering whether an optional section is something they want, then we should make the docs clearer about what this does, and the pros and cons of doing it. But if we don't want to tell people how to disable PasswordAuthentication, we might as well remove the section entirely, and I don't think that would be best.

xpucuto:
If English is not your native language you can get mislead by the headline.  I did read everything and i wanted to enter without a password . But finally i locked out myself and i couldn't log into the pi at all. And then you suggested me to turn that to yes . Anyway you know better than me.

scottleibrand:
Ok, so sounds like we need to make it a lot clearer that this is really optional and not necessary, and might be harmful if it doesn't work quite right. Would you like to add some more understandable text along those lines like you did for clarifying the rest of it?



On Fri, May 6, 2016 at 8:24 AM -0700, "Hristina Dimova" <notifications@github.com> wrote:












If English is not your native language you can get mislead by the headline.  I did read everything and i wanted to enter without a password . But finally i locked out myself and i couldn't log into the pi at all. And then you suggested me to turn that to yes . Anyway you know better than me.

xpucuto:
i would love you but i am afraid i don't understand this feature. All i know now is that i have to be careful with it. :)

scottleibrand:
Fair enough. Let's leave this open until one of us comes up with better text.

xpucuto:
Deal

danamlewis:
Closing because #272

PR #173, opened 2016-05-19
Update Configure-Automatic-Sensitivity-Mode.md

Spazholio:
Just general cleanup, and adding some commands to create the requisite reports.


Spazholio:
I need to go back and clean up a lot of the "[Your Path]" entries, but it's late so I'll tackle that tomorrow.

PR #183, opened 2016-05-22
New reference docs

Spazholio:
I'd love for people to take a look, and tell me what they think about the format.  Should anything be added?  I plan to add devices as well as reports - probably tomorrow or the next day.  And as well as a "Dependencies" section, I was considering going the other way as well - each page would have a "What reports depend on this report".  Shouldn't take too much doing.

And on that note, I'd like an opinion on something that naturally occurred to me after I had written 10+ pages.  Check this page in particular out: https://github.com/Spazholio/docs/blob/dev/docs/reference/openaps/openaps-report-settings-profile.md.  While I like having a "Dependencies" section to clearly indicate what is needed to make a given report run, what do you all think about just linking the filenames inline in the "Setup code" section?  I can think of pros and cons for each.  Thoughts?

Also, since all the files are in the same dir as one another, I *think* I got the internal linking correct.  I mean, they worked while I was testing them, so they should work when published, right?


danamlewis:
This is looking good @Spazholio! @scottleibrand already merged it, so keep 'em coming :)

(One nitpicky note in the sample you linked to on your version - I would prefer we list the maxiob as 0 in examples in the docs, so people don't blindly paste in a much higher value, and otherwise work through the setup process that starts with 0 and later encourages them to swap it out for a thoughtful value after they've figured out what is a good value for them)

PR #182, opened 2016-05-21
inserted information about upgrading the nightscout branch if necessary

MosiGitHub:
@logichammer asked for inserting some information on how to update Nightscout to the dev branch. I think that this step belongs to the chapter visualization and monitoring and added a sentence here. If you are just changing the branch of Nightscout you shoul be familiar with the necessary steps. That is why I decided to make just a brief note.

PR #184, opened 2016-05-22
upgrading the nightscout branch if necessary

scottleibrand:
Retargeting #182


MosiGitHub:
Thanks 👍

Gesendet mit BlueMail

PR #192, opened 2016-05-25
Update visualization.md

loudestnoise:
First PR woot woot. I think I'm doing this right. Adding my notes about where I think there's a disconnect in the docs in Phase 1


scottleibrand:
Awesome, thanks. In addition to the call out of what needs to be fixed, you linked people to a workaround. Perfect. :-)

PR #195, opened 2016-05-26
Update loops-in-progress.md

garykidd:
Added my name


scottleibrand:
We need to update the first-PR docs to get people to update their own forks before making edits, to minimize merge conflicts like this.

PR #206, opened 2016-05-31
Update understand-this-guide.md

val-albany:
Added a brief introduction of what OpenAPS is doing with all these commands ; )  Feel free to correct/adjust/add to it, just felt that knowing what's going on as early as possible was helpful.


scottleibrand:
Nice. Thanks!

PR #207, opened 2016-05-31
For Canadian / UK Users

apabari:
Correcting minor type-o in row 66


scottleibrand:
Guess I can consider this a 👍  for #220 and vice versa.  ;-)

PR #220, opened 2016-06-03
Update insulin_sensitivities conversion from RAW

thomasjacquin:
Added a <use> for Insulin Sensitivity unit converstion from RAW file. 
Prefixed insulin_sensitivities_raw.json with "settings/".


scottleibrand:
Guess I can consider this a 👍  for #207 and vice versa.  ;-)

PR #134, opened 2016-04-29
understand before doing

bewest:
Move the openaps mechanics/introduction ahead of the walkthrough.
According to @ddaniels1, this was useful.

PR #223, opened 2016-06-04
various updates to documentation based on our 'newbie' experiences

japrobo:
adding our name to the list of loops in progress


docs/docs/walkthrough/phase-0/hardware.md
```diff
@@ -22,7 +22,7 @@ If you're interested in working on communication for another pump (Omnipod, Anim
 		 * If you have one of the above mentioned pumps, but it has buttons that do not work, use the instructions found on this [Imgur photo album](http://imgur.com/a/iOXAP) to repair your pump.
 
 * <b>A way to communicate with the pump</b>:
-	* Medtronic CareLink USB stick is the recommended option for your initial loop setup
+	* Medtronic CareLink USB stick is the recommended option for your initial loop setup.  You can find this on the Medtronic website (https://medtronicdiabetes.secure.force.com/store/remotes-parts/carelink-usb-device/usb-wireless-upload-device)
```
danamlewis: Wouldn't include a link here, it's already included further down on the page where it breaks down more details on each piece of hardware. But, if there's a more thoughtful way to organize this page overall, would love to hear it!

docs/docs/walkthrough/phase-1/visualization.md
```diff
@@ -19,6 +19,142 @@ For Azure users, here is what these configuration changes will look like (with j
 
 Next, on your Nightscout website, go to the Settings (3 horizontal bars) in the upper right corner.  At the very bottom of the Settings menu, in the "About" section, you may check the Nightscout version (e.g. version 0.9.0-dev). Just above is a list of Plugins available.  OpenAPS should show up. Click the check box to enable. Similarly, in the case you've enabled the "pump" plugin, "Pump" should also show up in the list, and you may check the box to enable. You should now see the OpenAPS pill box (and any optional pump monitoring pill boxes) on the left side of the Nightscout page near the time. You may also want to graphically show the basal rates: select "Default" or "Icicle" from the "Render Basal" pull-down menu in the Settings. 
 
+# Configuring and Learning to Use openaps Tools
```
danamlewis: I see why you moved this here, but at this point I'd rather do two things: link over to using openaps tools and specify here that people need to go do that; and add a link at the bottom of "using openaps tools" that then links back here for finishing NS. 
Otherwise, we'd need to completely reorganize phase 1 (which you can argue for doing so), I think the "using openaps tools" needs to be it's own page for ease of re-finding it in the future. Sticking it in the middle of NS doesn't make sense, so I'd rather re-write the NS stuff to be appropriately contextualized here.

danamlewis:
I made a couple of comments in the commits, so wouldn't merge this quite yet. Let me know what you think, especially around re-organizing the NS stuff. If you're willing to tackle writing the NS stuff to be appropriately contextualized (this is a huge pending need, FYI), that would be awesome. If not, I think the short term fix is what I suggested about linking in NS section to the 'using' section, AND adding a link in the bottom of the 'using' section to easily link people back to finish setting up NS, if that's where they are in the process.

docs/docs/walkthrough/phase-0/hardware.md
japrobo: yup, I agree with your comment.  Found it later in the text.

docs/docs/walkthrough/phase-1/visualization.md
japrobo: I really want to rework the whole thing but as we are not yet done, I don't know how it would look yet.  Need more learning on my part.  I agree it is a big need!  Rewriting the NS stuff seems appropriate.

danamlewis: Actually, you might be the best person. The rest of us are too far away from it, so if you'd be willing to give it a first stab...that would be awesome. I think what it needs is for you/someone to look at phase 1 and restructure it in order, we can break it up into as many pages as possible. Might need a prep NS, then a using openaps tools, then a finish NS section? Think about that and see if it seems more doable? The current page structure isn't the right one, so we can delete/move as needed, we just need someone to break it into chunks and take the first stab!

scottleibrand:
Just merged most of this.  The stuff that duplicated or conflicted with #134 etc. I didn't.  Gonna PR dev back into japrobo:dev now so @japrobo can re-apply the conflicting changes and do another PR for the remaining stuff we still want to change.

PR #227, opened 2016-06-05
Adding Square/Dual-Wave workaround

Spazholio:
@danamlewis - I think this is sufficiently "this works for me, but YMMV" but if not, just let me know/change it up.


scottleibrand:
:+1:

danamlewis:
A few edits to suggest! Consider this style of writing, which will enable others to add tips/input on this topic:

* **How do I incorporate the Square-Wave/Dual-Wave features of my Medtronic pump into my loop?** 

Currently (June 7, 2016), OpenAPS doesn't treat Square-Wave or Dual-Wave boluses like you'd expect - it doesn't recognize the insulin given until the Square/Dual-Wave completes, and then it registers all of the insulin given over time as a single large "normal" dose. In addition, while the Square/Dual-Wave is running, temp basals are disabled. So many people do not use square/dual-waves as they would before they begin looping. This leaves many people with the question about what to do with large protein or fat meals.

Paul (link to your preferred public profile, like I did my Twitter account, or not) tip:  I don't eat very many carbs, but I eat a lot of protein and fat. Whenever I eat a moderate-to-large amount of protein, the excess protein gets converted into glucose. Before looping, what I've found works well for me is to do a Square-Wave Bolus for 70% of the protein over 3-4 hours in order to cover the conversion. However, keep in mind the above note about the square/dual preventing the loop from working. So in order to duplicate the behavior of a Square-Wave bolus, I've used a number of advanced OpenAPS features to great effect. In Nightscout, I'll enter a "Temporary Target" entry in Care Portal, leave the top and bottom targets at the defaults (80 for both), and set the duration for 180 minutes. If I'm eating some carbs with the meal in addition to the protein, I enter the carbs into my pump and dose for those like I normally would (in addition to setting the "Temporary Targets”), which allows the Advanced Meal Assist feature to dose more insulin than it would without knowledge of the meal.

MosiGitHub:
Ceben80 and I were discussing this topic 2 weeks ago and finally came to the possibiltiy to try it with the temptargets feature. Glad that you are offering this solution as a nice workaround for handling the missing functionality of square- and dual-wave boluses  :+1:

PR #237, opened 2016-06-11
Moved using-openaps-tools from Phase 2 to Phase 1 of the walkthrough

dm61:


dm61:
Not necessary -  this and more has already been done in wip-understand-before-doing.

PR #265, opened 2016-06-23
change 'gather-profile' to 'get-profile'

gideonaa:
Seems like there is an inconsistency here, am I wrong?


scottleibrand:
Good catch, thanks.  Merged manually along with fixing all the other instances.

PR #266, opened 2016-06-23
Possible correction on preflight operator logic

gideonaa:
I could be totally wrong here, but it seems like we needed this change in the preflight alias example.  It looks like if the clock-related check worked, it  would have outputted "PREFLIGHT OK" before ever running the  `mm-stick warmup` and the `sudo oref0-reset-usb` script.  Am I missing something here?

### DIFF
--- a/docs/docs/walkthrough/phase-2/building-preflight-safety-checks.md
+++ b/docs/docs/walkthrough/phase-2/building-preflight-safety-checks.md
@@ -37,7 +37,7 @@ returns the current pump time stamp, such as "2016-01-09T10:47:56", if the syste
 Collecting all the error checking, a `preflight` alias could be defined as follows:
 
 ```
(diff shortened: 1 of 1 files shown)


scottleibrand:
No, the way this logic works breaks down like this:

Do all the stuff, and if it all succeeds, echo PREFLIGHT OK.
If not:
  Check that you can talk to the Carelink stick.  If not, reset USB.
  echo PREFLIGHT FAIL and sleep for 2 minutes (in case the failure was due to another rig talking to the pump)

scottleibrand:
Your logic would only check the Carelink stick if the preflight succeeded, which is impossible if the Carelink is malfunctioning and needs a USB reset.

gideonaa:
@scottleibrand yup.  I was looking at it with too little coffee in my system I guess. Sorry!

PR #267, opened 2016-06-23
Remove possibly redundant clause in loop alias

gideonaa:
Again, I could be completely missing something here but it seems like running `monitor-cgm` on the left hand side isn't actually doing anything.  It would get the latest glucose entries only to have them deleted in the `preflight` step and then `monitor-cgm` would be run again inside of `gather`.  It's harmless I guess but possibly confusing?  Or again maybe I'm missing something.


scottleibrand:
Yes, I think you're correct here.  We also need to fix the `|| echo No CGM data.` part: that should say something more generic like "Loop failure" or similar.  Want to make that change as well before I merge?

gideonaa:
@scottleibrand yes I was thinking that too.  Done!

Issue #189, opened 2016-05-23
Don't use a dev branch for docs


mddub:
I think we all share the goal of making it easier for more people to contribute to the docs. Removing the dev branch, and using only master, would do a lot to further this goal.

- **Immediate feedback after merging:** "Hey, my edit is on the live docs now! That was easy! I can do this!" (as opposed to, "Hmm, it says merged, but the docs look the same... something about a different branch... oh well, I guess I just don't get this Git thing yet. I sure have a lot to learn!")

- **Not discouraging people:** Let's say you see an opportunity for improvement while reading the docs, so you check out/update the repo and start working, only to find the issue was already fixed on dev. Well, that was a waste of time. The duplicated effort and extra reasoning about staging/live versions can be frustrating to would-be contributors. This isn't a hypothetical: I've personally had this happen a couple of times, and it really put me off from contributing.

- **Staging doesn’t make sense for docs:** Who’s “testing” the dev branch, and how? By definition, if you’re making use of the docs, you’re reading master. If dev is truly the latest documentation, why is the live version out of date?

- **Staging is extra reasoning for new contributors:** The dev/master distinction introduces new contributors to an idiosyncratic workflow which is not easy to reason about. Using dev to stage a release of software makes intuitive sense. Using dev to stage static documentation adds one more stumbling block for someone building an understanding of version control.

There are instances where branching off from master makes sense (e.g. to collect documentation for a new feature before it’s been released), but those are called feature branches. Their release need not be coupled to all the other changes in the next “version” of the docs.

danamlewis:
Thanks for starting this conversation @mddub. One point to add to the discussion and evaluation of this - many people look at master and get angry if it's wrong. If we kill dev, there's likely going to be even more things that are wrong in master. (In my experience, this has been one of the successes of using dev, because we can merge things in more quickly and have multiple people then work to improve what's broken but is better than being missing.) A simple fix for this could be improving the starter guide to the docs, so people don't expect perfection...or something else. Any thoughts on this aspect? Looking forward to the discussion on this topic.

mddub:
@danamlewis I think PR discussion is the right place for vetting non-trivial changes to the docs. Merge quickly and fix later is bound to result in inconsistencies.

And yeah, I completely agree about a caveat. If there's not already one, I would include a statement like:

OpenAPS is an open-source project created entirely by volunteers. As such, this documentation is a continuous work in progress maintained by the community. (That includes you!)

[...]

PR #269, opened 2016-06-27
Hardware updates I made

SirBinky:
Sorry about the 3 commits, I didn't realise github was saving each edit i did. I copied the description worried I'd lost it. 

I'm hoping these changes make it a bit clearer and I updated the hardware to Raspberry Pi 3. 

I wasn't sure about keeping the Rpi2 stuff in but I thought for simplicity to remove it.

### DIFF
--- a/docs/docs/introduction/understand-this-guide.md
+++ b/docs/docs/introduction/understand-this-guide.md
@@ -19,3 +19,11 @@ One helpful thing to do before starting any software work is to log your termina
 You won't see a full loop where you can just download the code, press a button, and have a live loop. There are many places where there are examples, and instructions, but you must do the work to understand how to communicate between devices and transfer data between reports and files. This is key for helping you understand what you are building and how it will work. 
 
 In some cases, the documentation needs to be built out further, with easier to understand language and more examples. However, there are a few things (like a full `cron` example) that are not included in this guide, and intentionally so in order to ensure that you have full intent and autonomy in building your system for yourself.
+
+### But wait - I need the "Dummy" version
+
+Well, actually, you don't.  If you can deal with diabetes, invoking shell scripts isn't all that big a deal.  But it will probably help to have some idea what is going on here.
+
+It may help to think of the OpenAPS setup as a tiny "diabetes brain" which is focused only on figuring out how much basal insulin you should be getting.  It does this by collecting all that background data we usually let our pumps deal with: what is your insulin sensitivity, your target BG, the duration of action for insulin, etc.  Then, it collects more immediate data, such as what is the current IOB, and basal rate, as well as checking out the CGM to see what your BG has been up to recently.  Then, it decides what should be changed (if anything) and tells your pump to go to a new temp basal rate, either higher or lower, depending on all the other factors.
+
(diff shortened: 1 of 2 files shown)


scottleibrand:
Not sure whether or not we want to recommend the Pi 3 by default. It has much higher power consumption, which is problematic for some people who want to use it for a mobile rig. Let's open this PR up for comments.

xpucuto:
If you add the PI3 in the docs , you have to know that the Tecknet battery can not charge trough when the PI3 is connected . This is the reason i bought the Zendure battery which works great.

loudestnoise:
My RPi rig is an RPi3. I use the [MoKo 9000mAh battery](https://amzn.com/B00FGN1OBU) and I get easily 12 to 14 hours of run time without losing power to the RPi. I would recommend changing the recommendation to the RPi3 because of the added benefit of built in WiFi/Bluetooth.

shgrady:
I think it might have been helpful to mention the trade-off. I bought a lot of the wrong stuff in the beginning, like a wifi dongle you don't need with 3 (which could be corrected at line 115 if you keep both), since 2 and 3 at the same price point, 3 "seemed" like a good idea. Can we keep both and explain the difference in power consumption, because by using the 3, neither a UPS set up, or one of the powerbanks I bought worked. Long way from rig number 2, but definitely going back to a Pi2 for it when I do. Even though the Tecknet appears to not be charging through, the minute you disconnect the Pi, it jumps to all 5 LEDs.

Spazholio:
I'd recommend removing any explicit references to Pi2 or Pi3, but have a comparison page that outlines the pros/cons of each.  If the code runs identically on each one (which I believe it does) then I think advocating one over the other will just lead to, "The docs say Pi3, but I only have a Pi2 - will this work on mine?" type questions.

C-Ville:
I think it really depends on the setup you want to use. The pi 3 is nice for BT and WiFi. If you want to use an ERF module or TI over serial, than you will not have BT on the pi3 as far as I know. Therefore the benefit of a pi3 is not that great for such a setup. As for reference design I think the pi3 is great.

SirBinky:
I've rewritten the file based on the comments above. Adding back the Pi 2 data and changing Rpi to neutral versions outside the specific mentions. Also i've added the plus and minus points for each model.

danamlewis:
I like this with the revised changes, 👍 if someone fixes the merge conflicts.

SirBinky:
Sorry what do you mean merge the conflicts? I thought the newest doc was good to merge?

loudestnoise:
Yeah, I'm with @SirBinky on this. What merge conflicts are there? I mean I see the Github message saying there are, but if you click on Files Changed and it's just Paul's text vs Master. Or is the conflict also talking about other branches? For example, I've updated this page from my forked master branch, but I was under the impression we were supposed to be PR into the dev branch for review and then those with write access would then merge into master.

SirBinky:
I thought we forked the dev and pr'd the master?

danamlewis:
It says "this branch has conflicts that must be resolved", so there's another PR that got pulled in that made changes that conflict with whatever's in this one. So someone has to look in the command line and compare to see which changes conflict. The content from here has a thumbs up to get merged in, it's likely a matter of a small change in the old one or this one but it has to be done from the command line. 

Historically, you were supposed to fork dev and PR into dev, then we'd move dev to master, but we're allowing PRs to master now. So fork master and PR to master moving forward, unless you have huge gargantuan changes, in which case I'd maybe target to dev still so we can work through those.

loudestnoise:
Yeah, I just read #189 and I see that's the new procedure. I like it!

PR #272, opened 2016-06-29
remove Disabling password login section as it's confusing to new users

scottleibrand:
Per https://github.com/openaps/docs/pull/151 and discussion at https://gitter.im/nightscout/intend-to-bolus?at=57732a6a97171715547e48d1, probably best to just remove the Disabling password login section as it's confusing to new users.  Any objections?


tghoward:
No objection here. I think that is a good idea. Never disabled pwd login myself.

PR #274, opened 2016-06-30
Adding RileyLink to the Glossary

Spazholio:


docs/docs/Resources/glossary.md
```diff
@@ -23,3 +23,5 @@
 <b>ISF</b> - insulin sensitivity factor - the amount of insulin that drops your BG by a certain amount mg/dl. Example: 1 u of insulin for 40 mg/dl
 
 <b>NS, or Nightscout</b> - a cloud-based visualization and remote-monitoring tool. 
+
+<b>RileyLink (RL)</b> - A custom designed Bluetooth Smart (BLE) to 916MHz module - it can be used to bridge any BLE capable smartphone to the world of 916Mhz based devices. This device is focused on talking to Medtronic insulin pumps and sensors.
```
scottleibrand: I would say 900MHz instead of 916, as it works with 868MHz pumps too.

Spazholio: I took the info straight from https://github.com/ps2/rileylink as I couldn't think of a good blurb for it.  I can update it if you think it should be switched to 900MHz.

cjo20: "sub-1GHz" may be better, as "900MHz" doesnt necessarily seem to match 868 or 916MHz? Unless you're aware of devices that use sub-1GHz bands other than 916 and 868.

scottleibrand: Or maybe ~900MHz ISM band. @ps2, what's your preference? I agree that having this and the RL README match is good, but perhaps both should be updated.

danamlewis:
::pings @ps2 for clarification on the above::

ps2:
I would lean towards "Sub-1 GHz" if we want to imply more bands than 916.  I've used the RL successfully to receive omnipod packets at 433MHz.  And yes, updating the RL readme is probably in order as well.  I'll take care of that.

scottleibrand:
@spazholio want to make that edit here so we can merge?

ps2:
Just updated the RL Readme

PR #278, opened 2016-07-02
Update rpi.md

grhulz:
Perhaps this is obvious, but I ran the first command on my local computer and attempted to run the following commands on my pi. Perhaps this still isn't clear because the command to create a new folder needs to be run on the pi, but it should help clarify.

### DIFF
--- a/docs/docs/walkthrough/phase-0/rpi.md
+++ b/docs/docs/walkthrough/phase-0/rpi.md
@@ -196,19 +196,21 @@ Now you should be able to log in without a password. Try to SSH into the RPi2 ag
 
 
 ### Mac and Linux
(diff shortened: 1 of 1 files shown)


docs/docs/walkthrough/phase-0/rpi.md
```diff
@@ -196,7 +196,7 @@ Now you should be able to log in without a password. Try to SSH into the RPi2 ag
 
 
 ### Mac and Linux
-If you don't already have an ssh key, then on your local computer (*not* on the Pi), run `ssh-keygen` (keep hitting enter to accept all the defaults).
+If you don't already have an ssh key, then on your local computer (please note that all of the following commands in this section will be run on your local computer, *not* on the Pi), run `ssh-keygen` (keep hitting enter to accept all the defaults).
```
scottleibrand: Maybe would be better to just specify for each command which system to run it on, since `mkdir .ssh` has to be run on the Pi...

PR #280, opened 2016-07-03
update rpi.md

grhulz:
Silly mistake. I think it looks better now. I also added to the mkdir command.


scottleibrand:
Thanks. In future, you don't have to open new PRs when you fix something: the new commits automatically update the PR.

grhulz:
Scott,

Thanks. I was wondering about that. Sorry, I’m new to git, but I will get the hang of it.

Greg

PR #294, opened 2016-07-13
specify preferences

julieraines:
for the non-coders out there (we need all the help we can get) giving specifics on how to create a JSON or how to copy from Github


danamlewis:
This looks good! I actually would love to change all the examples where it says 2 to 0, for a further safety reinforcement...want to make those changes, and then I'll merge? Or if you don't have time, let me know, and I'll merge then go back and change.

PR #309, opened 2016-07-17
Update installing.md

julieraines:
So this would come after OVERVIEW and would have CREATE FIRST WORKSPACE at the end


scottleibrand:
Want to PR these to dev instead of to master, and then we can merge to master once all the moves are complete?

scottleibrand:
Retargeted against dev; closing.

PR #310, opened 2016-07-17
connecting to wifi and updating nightscout

julieraines:
I had trouble getting onto the dev branch of nightscout and Chris Oattes gave me some command lines that worked beautifully.  The update to connecting to wifi was a question I answered for Berat Kurar and now I see it's already been updated so ignore that part.


scottleibrand:
If you merge https://github.com/julieraines/docs/pull/1 that should remove the wifi part from the diff.

PR #315, opened 2016-07-21
Dev

gclyon:
Just trying to add myself to the loop!  Unclear as to which branch to use, but dev seems to be the one?


scottleibrand:
Good guess.  We're actually doing these little changes directly to master, though, and saving dev for restructuring and other stuff that shouldn't go in right away.  I retargeted and merged to master, so closing this one.

gclyon:
Tried master first, but never saw it in the pull requests listing. Might try it once more to make sure I'm understanding the process.

PR #322, opened 2016-07-23
Update .gitignore

pepecb:


scottleibrand:
Why? What is the current .gitignore doing incorrectly, or what does this fix?

PR #338, opened 2016-08-06
file phase-3/Open Loop Testing Phase: Tips & Lessons Learned gives problems with Git Gui on Win32 systems

PieterGit:
using : and & in filenames is not recommended


scottleibrand:
Why not rename instead of deleting?

PieterGit:
i can't rename this on a win32 system, the file is shown as deleted in my Git Gui. Probably due to the :  in the filename. can you rename it for me? I thought the file was stale and giving problems on windows.
It keeps coming back as delted because windows cannot cope with a :  in a filename.

I think this is a best practice: 
https://support.apple.com/en-us/HT202808
You should avoid using colons and slashes in the names of files and folders because some operating systems and drive formats use these characters as directory separators. Consider substituting an underline (_) or dash (-) where would normally like to use a slash or colon in a filename.

scottleibrand:
Is is used (linked) anywhere? Have you tried doing it through the Github GUI, or on your Pi?

PieterGit:
i used git-gui https://git-scm.com/docs/git-gui (this is not GitHub Gui btw). 
git-gui version 0.19.GITGUI
git version 1.9.5.msysgit.0

but I think other windows git clients also will have problems with : (colon) in the filename.
you can't write those files on disk with Windows (because it's resereved for the drive indicator, e.g. C:\path )

PR #345, opened 2016-08-08
Update using-openaps-tools (drop G4 with Share)

dduugg:
I don't think G4 with Share is supported with current versions of Raspbian/BlueZ based on https://github.com/openaps/openxshareble/issues/5 and debugging in person with @bewest (please correct me if I'm mistaken).


scottleibrand:
You may need to install a different version of BlueZ to get this working, but openxshareble definitely supports G4 with Share. (That's the only thing it supports in fact.) I would rather add the missing BlueZ steps to the instructions than remove them entirely.

dduugg:
@scottleibrand apologies for letting this PR lapse – i've updated it to fix the install command and linked to the issue tracking bluetooth compatibility issues.

PR #351, opened 2016-08-11
Update loop-and-retry-logic.md

dduugg:
Add a note to make the scripts executable. (Also formats the scripts themselves for readability)


scottleibrand:
We should probably move these scripts into oref0.

PR #369, opened 2016-08-27
Links and Formatting!

bfaloona:
* Markdown links displayed on readthedocs.io need a path, so `(page.md)` becomes `(./page.md)`
* Fix multiline source formatting of sample contents


danamlewis:
LGTM, but any reason not to target and put in master right away?

PR #379, opened 2016-09-03
Update using-openaps-tools.md

garykidd:


garykidd:
Apologies for not providing actual instructions on how to do what I was suggesting.  It looks like someone's prior PR was also a comment and not meant to be committed to.

PR #386, opened 2016-09-12
Added note about meal device

sjolundjohn:
Added clarification that if you are getting carbs from pump directly, you do not need to call it from Nightscout.


docs/docs/walkthrough/phase-6/Configure-Automatic-Sensitivity-Mode-and-Advanced-Meal-Assist.md
```diff
@@ -197,6 +197,7 @@ fields = pumphistory profile clock glucose basal carbs
 cmd = oref0
 args = meal
 
+* Only include carbs if you are getting this from Nightscout. If you enter carbs into your pump via the Bolus wizard, you can ignore add this to fields.
```
scottleibrand: Grammar nit.  Maybe "you can skip adding this field"?

PR #393, opened 2016-09-28
oref0-setup docs

scottleibrand:
This isn't perfect, but it's a lot easier to follow than the current docs, so I think it's ready for a quick once-over, one round of tweaks, and then a merge to dev.  I added the oref0-setup branch to RTD, so you can browse from http://openaps.readthedocs.io/en/oref0-setup/index.html


davidkeddydb:
Well when I first go to the Crontab I set them all at *5/ which I think was a mistake.  I guess a question or suggestion would be how much time does it to take to repower the usbs and for the devices to reset.  I also think that this is set for a plugged in Dex but not a nighscout Dex with sharebridge.  The cron jobs also hang after recycling through the Usb's.  Not sure if there is a programming issue with that crontab line of code after reset.  Also as there isn't an openaps do-everything I can't see where it might be failing.  Will try that off the alias that is written for it.
Get errors when running the aliases. specifically ns-loop:
s://JSON/shell/cgm/ns-glucose.json
json: error: input is not JSON: Unexpected '<' at line 1, column 1:
        <!DOCTYPE html>
        ^
json: error: input is not JSON: Unexpected '<' at line 1, column 1:
        <!DOCTYPE html>
        ^
ValueError: No JSON object could be decoded
Refreshed temptargetsCould not parse input data:  [SyntaxError: Unexpected end of input]
 and meal-carbs
Uploaded
pi@raspberrypi:~/dke2 $ openaps autosens
Wed 28 Sep 21:45:38 EDT 2016
detect-sensitivity://text/shell/settings/autosens.json
Could not parse input data:  [SyntaxError: /home/pi/dke2/settings/profile.json: Unexpected end of input]
reporting settings/autosens.json
I am still working on getting data to upload to Nighscout as that is not working

scottleibrand:
Following up with @davidkeddydb at https://gitter.im/nightscout/intend-to-bolus?at=57ec7b51465e867e6278afe5

davidkeddydb:
I think the sys log, tail -f /var/log/syslog should also be one to view on set up as I don't believe the tail -F /var/log/openaps/pump-loop.log show when the Carelink stick is faultering.

scottleibrand:
I find syslog to be way too verbose and full of false-positive harmless warnings to be useful as a primary method of monitoring the system. It would be good to add that (and methods like pausing the loop and running commands manually) to a troubleshooting section for how to investigate when it doesn't Just Work. Care to draft and PR anything?

davidkeddydb:
Yes,

Let me work on something.  I just can see people pulling their hair out when it appears the loop is running but nothing is happening,

I can definitely begin to draft a troubleshooting section and PR it.  I know Eric Jenson can likely add some good stuff to that but I will give a shot to starting it.


David Eddy
President
NEFCO Fire Investigations
www.nefco.net
(800) 675-8500

Submit an assignment<http://extranet.nefco.net/client.php>

PR #397, opened 2016-09-29
Dev

scottleibrand:
The reference/openaps links are still broken, but I can't figure out how to fix them in a way that's compatible with both GitHub and RTD.  Since https://mkdocs.readthedocs.io/en/0.9/user-guide/writing-your-docs/ claims the way that we're doing it should work, and since I'm not sure anyone uses those links much, I'm going to leave them broken until/unless someone comes up with a good way to fix them.


scottleibrand:
But all of the other changes here look good to me, so I think this is better than what's in master and ready to merge, so we can version it before merging oref0-setup to dev.

PR #401, opened 2016-10-06
Merge oref0-setup instructions to master (from dev)

scottleibrand:
I think we're about ready to start making oref0-setup the default.  It still has some rough edges, but it's *way* easier to use (and support) than manual builds. 

Anything we need to fix up in dev before we merge to master?


betluis:
I had already some time trying to understand the first instructions when you suggested going through the oref0-set up... so I was exposed to both. My comments so far:
first instructions help you understand the system from a programming point of view, while this could be good for some people for other could be extremely rough I think. The oref0 definitely helps one focus on the actual diabetes care concepts more quickly and with less distractions. I think this is definitely a right move although definitely there are clarifications to do.
As I move on through the process. I'll comment back later on again.

davidkeddydb:
Definitely update it to the new set up.  Way easier.

danamlewis:
+1 for @scottleibrand (resolving loop list conflicts), given feedback in Gitter & here, for going ahead and merging to master. If anyone looks at this in the future, please note that the "manual setup" instructions are still part of the docs, but have been moved downstream as part of the manual setup section.

PR #412, opened 2016-10-15
Update rpi.md

mccgm:


scottleibrand:
Does anyone have any suggestions for how to accomplish this?

vincepf:
Perhaps a reference to or copied text from "man 5 interfaces" on the Raspian Jessie? For example: "The manual page for the interfaces file explains the 'manual' option as follows:
`This method may be used to define interfaces for which no configuration is done by default. Such interfaces can be configured  manually  by
       means of up and down commands or /etc/network/if-*.d scripts.`
and the 'dhcp':
`This  method  may  be  used  to obtain an address via DHCP with any of the tools: dhclient, pump, udhcpc, dhcpcd. (They have been listed in
       their order of precedence.) If you have a complicated DHCP setup you should note that some of these clients  use  their  own  configuration
       files and do not obtain their configuration information via ifup.`
For more information, see `man 5 interfaces`"

Or, more simply "For more information see `man 5 interfaces`". 

There are many details that could be explained here about how the Linux networking scripts operate, but discretion should be used, as the purpose of the walkthrough is not a Linux class, but how to get openAPS working.

danamlewis:
@vincepf sounds like a good solution to me, can you PR that in? (feel free to open a separate PR to target openaps:master, and when we merge can comment here to link to that as resolution)

vincepf:
Added PR #513 with changes per above.

scottleibrand:
Fixed with #513 I believe.  Please open a new PR if you have any other suggestions.

PR #428, opened 2016-10-20
Dev to Master - Hardware Phase 0 changes

danamlewis:
Lots of changes to hardware section/phase 0 structure, but 95% same content. Here's the outline of what Phase 0 will look like with my proposed changes:

- Basic setup info
- Baseline data
- Hardware overview
-         Information about compatible pumps
-         Information about compatible CGMs
-         Hardware information related to Pi setups
-         Hardware information related to Edison setups
- Setting up Your Raspberry Pi
- Setting up Your Intel Edison*
- Loops in Progress

*This is the one new content page; pulled in info from mmeowlink wiki.


taylorfowler:
In raspberry-pi.md:

- Provide LiPo battery lInks for US / EU (RobotShop covers both i think: http://www.robotshop.com/en/37v-2000mah-5c-lipo-battery.html) 

-Can you make the "Next Step" link hit the "Setting up Your Intel Edison" page? or are those hardwired to the next file in the list.

In setup-edison.md:

-It's not particularly clear that in windows you need to open a command prompt in the folder you extracted to, (Shift-right-click --> Open Command window here), then just type "flashall" to run the batchfile. I think @diabeticsgonewild just uses cygwin to do her install? so this isnt necessary if you go that route (I didn't)

- for #4, I had to extract dfu-util for windows and drop it in the extracted jubilinux directory to get it to flash correctly (http://dfu-util.sourceforge.net/releases/dfu-util-0.9-win64.zip). I dont know if anyone else had that issue. (same cygwin-install-does-not-apply-here is probably true)

- that being said, it might make sense to do a quick bit on installing cyg-win and going that route may make sense. I was able to poke my way through the windows caveats just fine, buuuuut maybe not everyone can come to those conclusions based on the errors i was getting.

- More windows stuffs: Maybe include some links for Using Putty for SSH, and links to the GEtting started with Edison pages for connecting via USB Serial and over Wifi

danamlewis:
@taylorfowler - great feedback, thanks! I updated the battery link, and added a hard link in the edison section to help people jump to the setting up Edison section next, rather than the in-order-flow-next-link that goes to Pi setup. But if someone misses it, not a big deal, they should get there either way.

For your other suggestions, I'm going to suggest that after this PR is accepted that you fork a copy of master, make some of those changes, and submit a PR so that setup instructions don't get mixed in with this mostly re-structure PR. Thanks again!

taylorfowler:
@danamlewis No problem, will do!

audiefile:
:+1: I like this new structure a lot

scottleibrand:
http://openaps.readthedocs.io/en/dev/docs/walkthrough/phase-0/hardware/hardware.html links to http://openaps.readthedocs.io/en/dev/docs/walkthrough/phase-0/hardware/docs/walkthrough/phase-0/hardware/Can_I_close_the_loop_with_this_pump_May_20_2016.jpg which doesn't exist.  Looks like "/walkthrough/phase-0/hardware/docs/" may be extra?

danamlewis:
@scottleibrand resolved.

scottleibrand:
There's a URL on http://openaps.readthedocs.io/en/dev/docs/walkthrough/phase-0/hardware/edison.html that didn't format as a link due to a syntax error.

scottleibrand:
I don't believe it's possible to use MMCommander with oref0 at this time, so I would only link to subg_rfspy at http://openaps.readthedocs.io/en/dev/docs/walkthrough/phase-0/hardware/edison.html

danamlewis:
Ok - all links and notes from this fixed.

PR #431, opened 2016-10-22
setup script details

tghoward:
I really just wanted to get in the detail on the port for the explorer board, so the other text seemed to follow.


danamlewis:
Awesome, @tghoward

PR #445, opened 2016-11-12
List the dependencies explicitly

HebaruSan:
Some readers may be averse to running code from a URL, so provide the list of dependencies so they can be installed manually if desired.


danamlewis:
I like where you are going with this (and thanks for the PR!), but I'd like to put the automated version first, otherwise those less familiar will attempt to run things individually and get stuck. For those who don't want the automated version, having the split out steps below will be a logical flow, I think.

HebaruSan:
Thanks for taking a look! And I agree, the most commonly used section should be first. I've committed the change.

danamlewis:
Looks great, thanks!

PR #448, opened 2016-11-18
Updated 'Initial Setup' w/ PC vs. Mac instructions

osheroffs:
Added 'If you're using a Windows PC' header to existing section
Added 'NOTE: This command will not produce an output' to existing section

Added 'If you're using a Mac:' section [by copying the existing section]
Changed 'nano' to 'vi'
Added 'Type 'i' to get into INSERT mode'
Added 'NOTE: This command will not produce an output' to existing section

### DIFF
--- a/docs/docs/walkthrough/phase-0/setup-edison.md
+++ b/docs/docs/walkthrough/phase-0/setup-edison.md
@@ -54,6 +54,8 @@ If you have any difficulty with flashing, skip down to [Troubleshooting](https:/
 
 Log in as root/edison via serial console.
 
+If you're using a Windows PC:
+
     echo FIXME-thehostname-you-want > /etc/hostname
 
     nano /etc/hosts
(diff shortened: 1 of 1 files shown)


b-long:
Is this right?  `sed` on macOS has some minor differences with GNU `sed`, for instance a backup file extension is required.  For more info, see: [this discussion.](http://stackoverflow.com/q/4247068/320399)  Personally, I recommend this solution to the `sed -i` behavior: http://stackoverflow.com/a/4247333/320399 .

Additionally, it seems odd that you're referencing `/etc/apt/sources.list` on macOS.  Is this running in a VM?

scottleibrand:
The sed line gets run on the Pi/Edison, not the PC/Mac. The distinction is just because of the difficulty of running nano in a console window.

b-long:
Ah, my mistake! Thanks for clarifying 👍

scottleibrand:
If you can think of any edits that would make that clearer (preferably while also making it more concise) happy to include them.

PR #470, opened 2016-12-10
Removal of Password for Sudo Commands

Bender1061:


docs/docs/walkthrough/phase-0/setup-edison.md
```diff
@@ -138,6 +138,16 @@ Next, copy your ssh key to the edison if appropriate (directions you can adapt a
 
 * Edit /etc/logrotate.conf and set the log rotation to `daily` from `weekly` and enable log compression by removing the hash on the #compress line, to reduce the probability of running out of disk space
 
+* Edit to remove the Need for entering the password when running Sudo Command
```
scottleibrand: Should we add something like "If you'll be running as the edison user instead of root:"?

Bender1061: isn't the default user edison? only time I ever login as root, is if I really have too. And if you are logged in as root, you don't need to use sudo anyways.

scottleibrand: For the Explorer board you have to do everything as root for libmraa to work right.

PR #471, opened 2016-12-10
Added another loop-in-progress, and corrected some RPi3 information, as well as added some systemd instructions

vincepf:


docs/docs/walkthrough/phase-0/rpi.md
```diff
@@ -372,3 +388,8 @@ Open your crontab for editing -
 Save the file, then restart -
 
 `sudo shutdown -r now`
+
+or
+
+`sudo systemctl reboot`
```
scottleibrand: Maybe just change both of these to `sudo reboot`?

vincepf: That would certainly work, but the Raspbian Jessie man page for reboot states `These are legacy commands available for compatibility only.` Going forward, the 'preferred' system controlling commands are going to be via systemd (i.e. systemctl).

scottleibrand: I highly doubt they'll ever get away with taking away the `reboot` command, and if they do, everyone will just alias it. ;-)

vincepf: Haha, I'm sure you're right. But if you already know the `reboot` command, you don't need this document to tell you how to reboot. And if you *don't* know the command, why teach new users a "legacy" one?

scottleibrand: I'd bet `reboot` outlives `systemctl`, despite what the systemd folks think.  ;-)

vincepf: Hey, it's not a Linux course, right?  I won't be offended if these changes are discarded as not helpful or distracting.

PR #482, opened 2016-12-14
Update setup-edison.md

acerrone:


acerrone:
Directions state Mac users to follow 1-5 but I believe it should say 3-5.

tatigger:
Yes, I believe that is correct, however reading through the steps with the understanding that you are using a different program, Jubilinux instead of the file downloaded from the intel page, was helpful to me. Maybe, "read through, and then follow steps 3 through 5"

scottleibrand:
Resolved via https://github.com/openaps/docs/issues/494

PR #491, opened 2016-12-16
Small doc updates: adding mmol/L for improved readabiliy for mmol users

PieterGit:
small updates, adding mmol/L for improved readabiliy for mmol users  …
- consistent writing of mg/dL
- added mmol/L equivalent
- removed 'in-development' because the additional features are also
available on the master branch

PR #492, opened 2016-12-16
small updates, adding mmol/L for improved readabiliy for mmol users

PieterGit:
small updates, adding mmol/L for improved readabiliy for mmol users  …
- consistent writing of mg/dL
- added mmol/L equivalent
- removed 'in-development' because the additional features are also
available on the master branch


PieterGit:
now pull request against dev

PieterGit:
i also added this sentence/warning with eating soon mode. "Because your blood glucose will go lower, it's a good practice to make sure your sensor is reliable and check you blood glucose before setting the temporary target. Otherwise you may cause a hypo. ".

danamlewis:
Per #491, merging this same set to dev as well so they're in line still.

PR #500, opened 2016-12-17
Update README.md

Aaybob:
The production documents are an early version of current doc status. That early version has paragraphs like "here we will put instructions".


danamlewis:
We changed the community policy a few months ago and have been targeting PRs to master, so master/latest is actually most up to date. Any PRs with major structural changes will target dev, but master will usually be more up to date, so I wouldn't recommend this addition at this time.

PR #501, opened 2016-12-17
Update edison.md

Aaybob:
Questions about antenna answered.


docs/docs/walkthrough/phase-0/hardware/edison.md
```diff
@@ -14,9 +14,11 @@ You can use any charger with a USB plug, including a wall power charger. The Exp
...
+Again, we recommend the Explorer Board with a built-in radio stick. The antenna on board is a strip of copper underneath the green outer coating. The antenna is labeled A1. It will have its maximum power at 868 MHz. We will be using 916.5 MHz. The antenna has a line across it at one point with a label that says "915". You cut through the outer coating and the copper on that line with an exacto knife. A single cut is pleanty but if the cut doesn't look clean you could make two cuts and then dig out the circumscribed piece and then reseal the copper with nail polish. Then the antenna will have maximum power near 916.5 MHz.
```
danamlewis: Nice addition about antenna. We do support WW usage though, which is the 868. Maybe edit slightly to say something about "The antenna defaults to the 868 MHz range, which is what WW pumps use. If you have a US pump, mmtune will run and tune to 916.5MHz. If you want to boost the range of your antenna (optional - default should give you half a dozen feet or more on average range), then.." (and leave the rest of what you have about cutting.)

PR #503, opened 2016-12-19
Cron updates, Unknown Pill status comments

Tilade:
Added comments around Cron and how it may impact users who use Openaps manually for testing or setting up reports.  Also added comments around the OpenAPS Pill plugin and added "unknown" status based on the nightscout glossary located here: https://github.com/openaps/docs/blob/master/docs/docs/Resources/glossary.md


docs/docs/walkthrough/phase-2/troubleshoot-oref0-setup.md
```diff
@@ -8,6 +8,16 @@ You won't hurt anything by running the script (step 2) multiple times, as long a
 
 Cron is the scheduler that runs the loop. I.e. this is the automation feature to automate your closed loop. If you're using a test pump, it's pretty safe to go ahead and automate your loop. But if you're not sure, you can always come back and do this later.
 
+In order to send commands using `openaps`, cron must be momentarly disabled to prevent duplicate access to local resources.  To check if cron is running use `crontab -e` or `crontab -l`.  If you see a file filled with content, chances are cron is enabled.
```
scottleibrand: If you want to momentarily disable cron'd commands, you can just `killall -g openaps; openaps whatever` to run the manual command. Disabling cron is really only necessary if you want to do a bunch of manual troubleshooting.

Tilade:
I think outlining more methods for troubleshooting is my purpose.   I found
it helpful in verifying RF communication with a bunch of read commands.  I
also live in a "loud" house with lots of 900+mhz devices.  So running
mmtine is something I do more often than others I think during
troubleshooting. Not sure everyone is aware of how to stop a cron'd task
once enabled. Either method,  I think relaying that point would be helpful.

scottleibrand:
Agreed. I wouldn't remove anything you wrote. If you'd like to add the killall option as well, that'd be ideal, or we can merge and do it afterward.

scottleibrand:
Merged, thanks!

PR #512, opened 2016-12-23
Latest raspian release note

Aaybob:
says SSH is now disabled by default

### DIFF
--- a/docs/docs/walkthrough/phase-0/rpi.md
+++ b/docs/docs/walkthrough/phase-0/rpi.md
@@ -26,7 +26,8 @@ For the Path 2 Headless install, you will need:
 
 ## Download and Install Raspbian Jessie
 
-Note: If you ordered the recommended CanaKit, your SD card will already come imaged.  However, if you don't already know whether it's Raspbian 8 Jessie or newer ([see below](#verify-your-raspbian-version)), just treat it as a blank SD card and download and install the latest version of Raspbian (currently version 8.0, codename Jessie).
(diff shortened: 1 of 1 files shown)


vincepf:
This concern is mentioned already in the document under 'Path 2:'  "As of 12/11/2016 the Raspberry Pi Foundation is disabling SSH by default in Raspbian as a security precaution. To enable SSH, create a file called ssh and save it to the boot directory of the mounted drive. The file can be blank, and it has no extensions. This will tell your Pi to enable SSH."

Aaybob:
I read the 'Path 2:' note in the documents just now.  Is path two the only one that lets one write a file to the root partition? The Readme file uses the word partition instead of directory. Does it mean "in the /root directory" with the files .bash_history, .bashrc, .npm, and .profile? My /root directory doesn't have the file but SSH is working.

vincepf:
The 2 paths are simply different ways of accomplishing the same things. If one is able to write such a file to the SD card prior to the first boot, then it is possible to set up the Pi without a monitor or keyboard attached (headless).

 The SD card will have 2 partitions, one of which is mounted by default as the /boot partition. It will be marked as bootable and read as such by another device (such as one might be using to write a 'ssh' file).  

On the Pi, after booting, the boot partition will be mounted as the /boot directory.

Directory and partition are not interchangeable terms, but in this case they mean the same thing. The walkthrough simply words it as if you are logged into the Pi ("the boot _directory_ of the _mounted_ drive")

The fact that you have ssh working with no ssh file only means you have enabled ssh after the initial boot, or it may have been enabled prior to your upgrade and defaulted to stay enabled during the upgrade.

Aaybob:
we found raspi-config must be run to enable ssh

scottleibrand:
Is this still needed, or have the other edits since then covered it?

PR #514, opened 2016-12-27
Edison module info

dougtakeuti:
It would be good to know that there is a lower power consumption (and heat dissipation) version of the Edison.


scottleibrand:
Updated with some clarifications.  Will wait for https://github.com/dougtakeuti/docs/pull/1 to be merged to verify the diff before approving and merging.

PR #522, opened 2017-01-04
Updated Mac details for accessing Edison console.

tim2000s:


docs/docs/walkthrough/phase-0/setup-edison.md
```diff
@@ -65,7 +65,7 @@ To flash the Edison using a Raspberry Pi, you’ll need a large (preferably 16GB
   - Once you've made those changes, Click on OPEN at the bottom of your Putty configuration wondow. You may need to click on Enter on your key board a few times. 
 
 ### If you're using a Mac:
-  - Open a terminal window and type `sudo screen tty.usbserial-* 115200` or similar.
+  - Open a terminal window and type `sudo screen /dev/tty.usbserial-* 115200` replacing the '*' with your Edison UART serial number, obtained using lsusb.
```
scottleibrand: Most people should be able to paste this command as-is, as long as they only have a single Edison connected...

tim2000s: I'm running on a MacBook with USB-C extender to 3 USB ports. If you run as is, it doesn't work.

scottleibrand: Without the /dev/ it definitely doesn't work, so we definitely need to fix that.  `sudo screen /dev/tty.usbserial-* 115200` works for me, though, without replacing the `*`.  Does that work for you with the USB hub?

tim2000s: Without the specific serial number, `sudo screen /dev/tty.usbserial-* 115200` doesn't work via the hub.

PR #523, opened 2017-01-05
Updated oref0-setup and BLE tethering iPhone and CareLink support

Aaybob:
I skipped MMeowlink  and never read the parenthesis and others have also done similarly. Text change to avoid wrong MMeowlink assumptions.


docs/docs/walkthrough/phase-2/oref0-setup.md
```diff
@@ -44,7 +44,7 @@ Pull/clone the latest oref0 master by running:
 
 ## Step 2: Run oref0-setup
 
-__Note:__ If you're using the 915MHz Explorer board, you'll need to log in as root to run oref0-setup.sh, as the mraa package doesn't yet support running under an ordinary user account.
+__Note:__ If you're using the 915MHz Explorer board, you'll typically log in as root, dependencies and git pulls logged in as edison is fine but for oref0-setup.sh the mraa package doesn't yet support running under an ordinary user account.
```
scottleibrand: I think "typically" is too weak. We should tell people to always run everything as root if using the Explorer board. The oref0-setup also needs to run as root, so that the loop gets scheduled in root's crontab and can run mraa properly.

Aaybob: My confusion. There was more to this PR. Happy to drop this particular edit from the PR
.

scottleibrand: That would be good. Do you know how to update your branch to change it back?

Aaybob:
I changed the Bluetooth tethering feature. Sorry that I'm putting all doc changes together, still learning how to PR. Let me know if you cannot access updates to Bluetooth tethering doc.

docs/docs/walkthrough/phase-2/oref0-setup.md
Aaybob: So lost...........

Aaybob: this pr is about carelink. git should move on from this resolved issue. but its not as smart as us.

scottleibrand:
Just put out a call for volunteers at https://gitter.im/nightscout/intend-to-bolus?at=5873fb41074f7be763ae11da - it would be good if we could get someone to update it to do the preflight check (and ( mm-stick warmup 2>&1 || sudo oref0-reset-usb ) stuff if it fails) automatically during oref0-setup, rather than making people do it manually when setting things up.

PR #527, opened 2017-01-06
Description of different Edison variants.

woodtrevor:
Added a brief description of different Edison variants/model numbers with some reference material and links to 2 online retailers that allow the buyer to choose which model to purchase. This may be useful as some users have reported heat issues with their Edison modules.


docs/docs/walkthrough/phase-0/hardware/edison.md
```diff
@@ -4,6 +4,10 @@
...
 
+Note: There are several different model numbers or variants of the Intel Edison Compute Module. This does not appear to be documented on Intel's website or at many of the online retailers. However, the different variants can be seen in the product listings at [Mouser](http://www.mouser.com/Embedded-Solutions/Computing/_/N-aez39?Keyword=intel+edison) and [Arrow](https://www.arrow.com/en/products/search?q=intel%20edison&filters=Manufacturer_name:Intel;).
+
+It appears as though the main differences pertain to onboard vs. external antenna and low power vs. high power wireless radio options. The low power radio variant is classified by Intel as "Wearable". According to some discussion on the Intel message boards (see [here](https://communities.intel.com/thread/81519) and [here](https://communities.intel.com/message/251806#251806)) the "Wearable" variant limits the wireless radio power to "keep the thermal properties at a lower level". Some users have noted that their Edison modules get very hot at times. The low power "Wearable" variant may be useful to avoid heat issues.
```
scottleibrand: Do we have any info indicating this is controlled in hardware vs. software?

woodtrevor: It's a good question. It seems clear that there are hardware variations. But information is sparse and it isn't really clear what the exact differences are. The implication is that the low power version is physically limited to reduce heat. I don't know how much control is possible via software. I did some more poking around but did not turn up anything definitive. The Edison WiFi guide does have a section about disabling power management but it doesn't explain how or why that would be useful. It shows how running iwconfig displays the transmit power. Would this differ on the low power vs. standard power variants? How would the actual wireless performance differ between them? Idk. It would be interesting to experiment on the two side-by-side. I tried to word the text so as not to imply a recommendation but just to point out that there are some differences. Maybe this just confuses issue...?

scottleibrand: I was thinking about adding something to the effect that "it's not clear whether or not different Edison models would use lower power in an OpenAPS configuration" or something like that.

woodtrevor: Yeah, that's a good idea.  How about: "It is not clear whether or not different Edison models would use lower power or generate less heat in an OpenAPS configuration."?

scottleibrand: :+1: If you can update your branch with that I think this will be ready to merge.

woodtrevor: I updated the text to make it clear that the different models have not been tested to confirm better performance by one compared to the others. Does that look better?

PR #535, opened 2017-01-09
Clarify Port options for mmeowlink

rdalisky:
Separate options for mmeowlink ports.


danamlewis:
This one is not quite right - you added about for enlite sensors. It doesn't matter which cgm you use. mmeowlink software is what you use for a non-carelink radio stick. Were you just trying to make the explorer board port more obvious; or is there something else you're trying to achieve with this PR?

rdalisky:
(Enlite sensors) should be removed from the top line in my PR.  i think separating the options into two lines will make it more apparent.

danamlewis:
Agreed. Want to make the change of removing Enlite sensors; or want me to do it as I merge?

rdalisky:
Commit added!

PR #541, opened 2017-01-12
Update raspberry-pi.md

Aaybob:
RPi2 RPi3 clarify the difference


docs/docs/walkthrough/phase-0/hardware/raspberry-pi.md
```diff
@@ -35,20 +35,19 @@ High level list of supplies needed for a Pi-based setup:
...
+
+For the more advanced RPi3 a larger power supply is suggested, 2.4Amps and 5.25Volts. We never want to under power computers but we want a convienent size to take the system mobile. When away from a single power supply we power the RPi2 with a battery that provides 2.1A, 5V and stores 10Ah, 10000mAh. Such batteries are adequate for the RPi3 running OpenAPs too. 
+
+RPi3 is a direct replacement for RPi2. RPi3 has an on board Edimax WiFi. The Edimax WiFi adapter is commonly sold with RPi2. If you use the Edimax WiFi adapter on a RPi2 then a fully programmed micoSD may be taken out of one version of RPi and successfully inserted into the other.  
 
-The Raspberry Pi 3 has higher specs and built-in WiFi and Bluetooth, so it draws
-more power. As a consequence, it has a shorter battery life than the
```
scottleibrand: This is an important point and should not be removed.

docs/docs/walkthrough/phase-0/hardware/raspberry-pi.md
```diff
@@ -35,20 +35,19 @@ High level list of supplies needed for a Pi-based setup:
...
-The Raspberry Pi 2 has fewer and lower spec components and so draws less
-power, but requires a WiFi adapter to be also purchased. The spec makes no
-difference to the OpenAPS app, so either model is suitable choice.
+## Raspberry Pi 2 and 3
 
-[Raspberry Pi 2 Model B](https://www.raspberrypi.org/products/raspberry-pi-2-model-b/)
+[Raspberry Pi 2 Model B](https://www.raspberrypi.org/products/raspberry-pi-2-model-b/) and [Raspberry Pi 3 Model B](https://www.raspberrypi.org/products/raspberry-pi-3-model-b/) are both excellent choices for your first implementation of OpenAPS. RPi3 is an improvement of RPi2.
 
-## Raspberry Pi 3 Model B  
+When handeling RPi2 it was found that the microSD could be inadvertantly dislodged crashing the system and sometimes corrupting the data and program present on the microSD. The RPi3 has fixed that design flaw. Another RPi3 improvement is better USB connector hardware in which the chord fits more tightly. This tighter fit provides a greater transmit range for the CareLink.
```
scottleibrand: I'm not clear how a tighter cord (not chord) fit provides greater range.

Aaybob: the high speed communication signal sees a discontinuity causing reflections noise on the signal. What I noticed was a much better usb fit and much better range so assumed they were connected.

scottleibrand: That doesn't feel like a strong enough evidence to support a hypothetical mechanism like that. I would remove the last sentence and just state that it has a tighter cord fit.

Aaybob: It could also be a result of more power being used. But the snug fit is a contributor.

scottleibrand: I also don't want to give anyone false hope about Carelink range. Once they get up and running, everyone should be moving to a TI stick and/or Edison+Explorer.

docs/docs/walkthrough/phase-0/hardware/raspberry-pi.md
Aaybob: RPi3 max power draw is 2.4A compared to RPi2 1.8A. I didn't like the lack of specificity. Do we have some numbers. Our application is not maxing the power usage. The explanation of higher specs, built in WiFi and Bluetooth is not satisfying. So it is hard to know how much shorter and do we want people to buy a 20Ah battery and a backpack?

scottleibrand: I like more specifics. Perhaps my real problem is with the nonspecific value judgement that "RPi3 is an improvement of RPi2." It is on many fronts, but if anyone wants to take their Pi mobile, the requirement for a bigger battery may outweigh all of that.

Aaybob: Its an improvement for a non mobile system in my experience. And we've changed our recomendation to using edison for mobile systems.

Aaybob: I'm not comfortable asserting a larger power draw. I don't have the setup to test that. I can report how many hours RPi3 runs with a hardwired Dexcom. Then we could compare that length of time to a recorded RPi2 experience. And state that for the record.

Aaybob: Let's do that. I'll fully charge a KMASHI 10000mAh Model K-MP816 and run RPi2, CareLink without extension and Dexcom connected with Dexcom chord and record the length of operation. And put that in the text for a reference run. And we can request similar trials with requests to PR them.

scottleibrand: To isolate the effect of the RPi2 vs. RPi3, we'd need to do the trials back to back with the same battery pack, as batteries' capacities gradually decline over time and may not be identical even when new. We'd also need to make sure that the Dexcom receiver is fully charged before unplugging the battery to start the test. 

Do you have both an RPi2 and RPi3?

Aaybob: no     I mistyped RPi2 where I meant RPi3.
it       surprises me how difficult this data is to come by
The PR now restates what we know and leaves out assumptions and value judgements.  It now states:
power the RPi2 with a battery that provides 2.1A, 5V and stores 10Ah, 10000mAh. The RPi3 is hotter than the RPi2 so a larger battery may be needed? The larger the battery storage capacity the longer the battery lasts.

PR #543, opened 2017-01-13
turning cron off and on

Aaybob:
Is what the section is about. I made the section more verbose.


docs/docs/walkthrough/phase-2/troubleshoot-oref0-setup.md
```diff
@@ -4,19 +4,19 @@
 
 You won't hurt anything by running the script (step 2) multiple times, as long as you name it something different. If you already have a working loop and are testing the setup scripts, just make sure to comment out in cron the loop you don't want running.
 
-## Should I enact cron?
+## Turning cron off and on
```
danamlewis: I do want to leave the initial header "Should I enact cron", but your new header goes with your new additions - so maybe revert the original title, and then add your new title above your new section that's specifically about turning it on/off.

PR #544, opened 2017-01-13
Update my-first-pr.md

Aaybob:
After some struggle lets make it easier I say.


Aaybob:
Yes, navigating is something I've been able to manage but may be difficult for others. The directory tree was at first incomprehensible. So I like your idea a lot. When I read the detailed breakdown I can't open the door to my brain. So separating the steps as icon clicks from the steps as git operations helps. Thus this new section followed by the existing.

Aaybob:
Oh wow I'll put that method in. I had never known what that box at the bottom right did?

Aaybob:
The change requested (navigate from readthedocs.io/en) was made. It is so easy for us now to PR but it used to be extreemly dificult for me. Removing technical terms may help. I did not remove the section with technical terms because I still don't know what it says.

PR #545, opened 2017-01-14
Update setup-edison.md

jaylagorio:
Updated the Windows PC page file section to reflect that it may not need to be done on machines with high specs and that failing to return the settings to their original values could cause system instability.


danamlewis:
@jaylagorio, do you mind retargeting to master?

jaylagorio:
I think that just did it, is that what you meant?

danamlewis:
@jaylagorio yep, perfect! thank you. merging now.

PR #549, opened 2017-01-15
Clarify that the paragraph on radio sticks doesn't apply to Explorer Board

j4cbo:
Clarify that the paragraph on radio sticks doesn't apply to Explorer Board. (edison.md)


danamlewis:
This looks good, but can you retarget to master? We're sending most small changes like this to master. Thanks!

PR #550, opened 2017-01-16
Clarify that the paragraph on radio sticks doesn't apply to Explorer Board.

j4cbo:
Clarify that the paragraph on radio sticks doesn't apply to Explorer Board. (edison.md)


danamlewis:
Thanks for retargeting! and for the PR. 👍

PR #555, opened 2017-01-17
Additional Carelink troubleshooting steps.

woodtrevor:
Added additional notes in Phase 2 troubleshooting regarding the Carelink.  Also added a link there to the Resources troubleshooting page that has an entire section on dealing with the Carelink.  Added additional text to that page to describe Carelink errors that need intervention and to suggest doing this automatically via a cron job.

I see this pull request also includes changes I made previously for a different pull request that have already been merged in.  Do I need to create a new fork before submitting a new pull request?  Or is there a way to exclude those changes from this pull request?


danamlewis:
I see those were already added to master, so not sure why it's re-including those commits, other than maybe you didn't merge openaps master back to your master before starting this PR? A new branch would solve it, but not a big deal here - if I wanted to I could cherry pick just the new commits, but I'll approve all and see what happens since we had already approved the previous changes about Edison variants. thanks for this PR!

Aaybob:
The variants have the same data sheet. There is no documented difference. It may be just marketing.

PR #557, opened 2017-01-21
Update setup-edison.md

Aaybob:
the flash process may require multiple attempts. This tells another way to know if success is achieved.


Aaybob:
The process that flash takes on windows to edison is:
flashall.bat, plug in only OTG, wait for completion and five minutes, unplug OTG to achieve no power, plug in UART to begin boot and PuTTY login. Then control D when/if asked.

PR #558, opened 2017-01-21
Added in section on Pancreabble use and alternative mechanism for resolving.conf

tim2000s:


scottleibrand:
I can't fix the merge conflicts here, so did so in https://github.com/openaps/docs/pull/596 instead.

PR #562, opened 2017-01-21
Update setup-edison.md

Aaybob:
if edison isn't flashed serial port doesn't work
The location of this paragraph implies otherwise.


Aaybob:
this ads this   - The Serial Port connection described in this section only functions after the Edison Compute Module has been flashed successfully.
and 
repeats that here:
 And if you're using an Edison Compute Module that has already been flashed successfully. 
    - Go to Control Panel\All Control Panel Items\Device Manager\Ports\ and look for USB Serial Port COMXX.  

of course its only confusing and needs this explanation when it's all brand new

scottleibrand:
You can connect via console regardless of the flash state of the Edison.  If it is not flashed, it will still show the bootup sequence over serial console after you power it on.

PR #582, opened 2017-01-26
Update setup-edison.md

apabari:
Added an explicit line item to install 'npm'


danamlewis:
Closing - because this section is now in phase 4, and you should be successfully looping and not needing npm re-installed at that point :) Thanks @apabari!

PR #584, opened 2017-01-27
Add xdripAPS offline monitoring info into docs

danamlewis:
Pulling in @colinlennon 's great WIP instructions about using xdripAPS. This will need some more testing.

### DIFF
--- a/docs/docs/walkthrough/phase-1/offline-looping-and-monitoring.md
+++ b/docs/docs/walkthrough/phase-1/offline-looping-and-monitoring.md
@@ -11,7 +11,7 @@ Dexcom CGM or other CGM users will need alternative solutions, which range from 
 ## Offline monitoring
 
 * See Pancreabble instructions below for connecting your rig to your watch
-* See xDrip instructions for seeing offline loop status (coming soon)
+* See xDrip instructions below for seeing offline loop status (very rough WIP - please PR any fixes/edits to it)
 
 ### Note about recovery from Camping Mode/Offline mode for Medtronic CGM users:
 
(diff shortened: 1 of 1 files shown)


chrisharris751:
One issue that xdrip G5 users will run into is xdrip resetting bluetooth when it has trouble connecting for a reading. This causes the bluetooth tether to drop and requires manual re connection. One solution is the collector running on an android wear watch with xdrip+ (this is my current setup and works great) and the other thing I have found that helps is an app called BT AutoTether. It tries to auto reconnect if the tether is dropped. Its not always successful but it helps.

colinlennon:
Thanks Dana! Yeah Chris, I use BT Auto Tether as well. Personally, I find that BT tethering is rock-solid - it always reconnects automatically, e.g. after I have moved out of range. In xDrip+, I have disabled the "G5 Bluetooth Watchdog" setting and that stops xDrip+ from resetting BT. The settings required vary a lot depending on phone model though. I'm not sure if we should document that type of thing in the OpenAPS docs - I'm just conscious that it doesn't relate directly to OpenAPS. The xDrip+ Gitter channel is probably the best place to get advice. Maybe I should add some info on my own personal setup to a new page in the xDripAPS repo and we could also add some details of the Android Wear approach too, and include a link to that page in the OpenAPS docs?

The Android Wear approach sounds great, which watch do you have? How long does the watch battery last between charges?

chrisharris751:
Colin, yeah it probably doesn't belong in the openaps docs. I was just sharing my feedback/experience. 
The watch I'm using is a Sony smartwatch 3. It's absolutely fantastic. With WiFi turned off I can easily go a full day with out a charge. I charge it while I'm in bed and it generally always makes in through the day. If it doesn't it charges via a micro USB port very quickly. So 15 mins of charge is all that would be required if you have a long day in the office.

I got mine off eBay for less than 200 dollars AU. At that price, since it waterproof I've been showering and swimming with mine with out any problems. I've even taken it into the surf! Cheers.

colinlennon:
Thanks again Chris - yeah, it'd be good to figure out a good way to capture different people's experiences and setups, as xDrip with G5 seems to be very unpredictable, depending on phone model, Android version, etc. and a lot of trial and error is needed at the moment. I know there's a spreadsheet in the G5 Facebook group, maybe I should link to that actually. Your watch setup sounds fantastic, I'm very tempted to get a Smartwatch 3! I use a Pebble at the moment and love being able to set temporary targets with it, via IFTTT. I'll need to do some Googling and figure out how that can be done from an Android wear watch.

(Sorry for going off-topic Dana!)

danamlewis:
No worries! Thanks for writing this up (and documenting these issues). I'm going to go ahead and merge this so people can PR on top of it, and +1 for adding future links/additional sections or pages as needed. Thx again!

PR #586, opened 2017-01-28
AMA Visualization

apabari:
Tweaked documentation for AMA forecast in NightScout


danamlewis:
Can you re target to master? (Should be able to tap something to change it). And first sentence seems like a fragment? Otherwise awesome, thanks!

PR #594, opened 2017-01-28
clarified some code shading

Kdisimone:


danamlewis:
@Kdisimone not sure about this one - it undoes a lot of the additions scott just provided. Maybe close, re-update your branch, and then try this one again?

PR #598, opened 2017-01-28
Fix photos that incorrectly showed yocto

walker0:
Fix photos in new mac setup guide that showed yocto instead of jubilinux


danamlewis:
heh - I was hoping someone would catch and swap those out :) Thank you!

PR #610, opened 2017-01-31
Creating a "accessing your rig" page for Phase 2

danamlewis:
Content from @Kdisimone


scottleibrand:
This is great stuff @Kdisimone - thanks!  You even taught me something that will be extremely useful: I didn't know about the iNet network scanner app.  😃

PR #638, opened 2017-02-06
Create how-to-update page

danamlewis:


dramageek:
Do we need to mention Cron at all? Not sure if the setup script properly overwrites old configs. I've always just hashed out the old stuff before I ran it.

PR #639, opened 2017-02-07
Drastically update understanding-this-guide

danamlewis:


bdr1177:
I like the changes to the high level explanation and the addition of the high level image. It will be helpful for novices starting out to have an understanding of how the parts of the project fit together. The only thing that may be confusing for people starting out is that it mentions Edison and RPi above, but the image and related text only refer to the Edison.

PR #643, opened 2017-02-08
Update loops-in-progress.md

ig888ack:
Working on my first loop!


danamlewis:
woohoo, another Alabama looper! @jyaw is in Hsv, fwiw. (and I’m from Hsv, and will actually be heading south through the state in April in case it works for a meetup)

jyaw:
Hey Kyle (@ig888ack)! I have a friend in Huntsville looping as well, so there's at least 3 of us. Cool to see this thing get used in our state. April it is! :smile:

PR #646, opened 2017-02-09
Update loops-in-progress.md

iainct:
Adding my name to the list, at place 47 as I was the 47th to get looping, however maybe I should have put it at the end...?


danamlewis:
(It really doesn't matter about order or anything - this list is meant to be a playground for people to practice putting in pull requests while they're working on building. The who-is-actually-looping list is in the wiki, in case you also want to add yourself there: https://github.com/openaps/docs/wiki/Where-in-the-world-is-%23OpenAPS%3F)

PR #659, opened 2017-02-13
Pr update with images

Kdisimone:
Updating PR process to be simplified with screenshots.  Deleted the second half to be clear that the PR was already created.  Added explanation that if starting PR requests from their own fork, they will need to perform a compare and update first.


danamlewis:
Beautiful. Thank you!

PR #660, opened 2017-02-13
Fixing links to images

Kdisimone:
Image links to my PR branch were not functional once my branch was deleted. Updated links to openaps doc's master branch where they now reside.  Hopefully the new image (PR7) will pull and the link will be correct for that new one without needing edit.


Kdisimone:
next time i'll just use the shortened pathway so that it shouldn't be an issue.  👍

PR #665, opened 2017-02-15
fixing photo links again?

Kdisimone:
Perhaps this will fix the photos not showing in the rtd version of the files


Aaybob:
The problem is more universal than a photo link issue. Other reference links are also non-functional even though they are all correct. Both photo and other links do function in edit preview mode. So I don't think this change is warranted.

PR #669, opened 2017-02-17
Updates

dramageek:
Change Vi usage to Nano to match the rest of the documentation. No sense in making newbies learn another editor.

Remove mention of 'weekly' setting in logrotate- Jubilinux sets it as default.

### DIFF
--- a/docs/docs/walkthrough/phase-0/setup-edison.md
+++ b/docs/docs/walkthrough/phase-0/setup-edison.md
@@ -119,9 +119,8 @@ Run these commands to set secure passwords:
   
 ## Multiple Wifi Networks:
 
-`vi /etc/network/interfaces`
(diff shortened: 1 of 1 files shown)


scottleibrand:
We very specifically changed the directions to use vi, as everyone was having trouble with line wrapping using nano over console. The directions say to change logrotate *to* daily *from* weekly (the default).

dramageek:
Hmm. I was wondering if there was some Nano weirdness that Vi avoided. I'd
run into odd glitches with Nano via console on Mac.

I can't believe I misread the other one. I bet I've flashed my Edisons ten
times, and misread it every time!

scottleibrand:
If the logrotate directions are unclear, a PR to clarify might be worthwhile.

PR #670, opened 2017-02-19
Add alias instructions

Kdisimone:
Adding it as a new page for Phase 4.  Not sure if that's the appropriate place, but seems about right


danamlewis:
Let's stick this one in Phase 1 again, and also add .md to this one's name, then will merge. Thanks!

Kdisimone:
Ok, should be good to go now.

PR #676, opened 2017-02-22
autosens_max and autosens_min

Kdisimone:
updating for autosens_max and autosens_min being respected by autotune.  also add note about mid-day pump basal changes while autotune is enabled in oref0 loop.


Kdisimone:
Looks great.  thanks

Kdisimone:
A couple questions...

scottleibrand:
You can make additional commits to the branch that will be automatically added to the PR. 

Autotune only does 24h by default for each run, but every run uses the previous run as its starting point, so the iterative nature means the autotune results are based on weeks' worth of data. 

I've never really played with the excel export myself.

PR #679, opened 2017-03-01
Clarifying some instructions

Kdisimone:
adding a couple easy troubleshooting, such as for "can't find a PTY" in screen mode.  Removed the ping google.com command because that step right before is using ssh to login.  Would be redundant to run ping if successful at getting in through ssh.  If we want to keep the ping google.com, then we would need to add a line to state control-c to exit the log


docs/docs/walkthrough/phase-0/edison-explorer-board-Mac.md
```diff
@@ -260,20 +262,20 @@ These are the wifi networks that your rig will be able to use to stay connected
...
-* Wait as many lines of action go by in the Terminal window...eventually you will get to a prompt.
+* Wait as many lines of action go by in the Terminal window...eventually you will get to a prompt that has your new edisonhost name login.  We aren't going to login right now.  Just saving that window in case we need it later.
 * Open a new Terminal window by pressing Command-N
-* Login to your Edison by entering `ssh root@edisonhost.local` (changing edisonhost to the hostname you selected above)
+* Login to your Edison by entering `ssh root@edisonhost.local` (changing edisonhost to the hostname you selected earlier above)
 * Enter your password that you set earlier
 
 ![Login to your rig](../../Images/Edison/Rig_login_time.png)
 
-* Run `ping google.com` to make sure your rig is online.  If not, go back and check your /etc/network/interfaces and /etc/wpa_supplicant/wpa_supplicant.conf files above: you probably either missed a step or made a typo.
```
scottleibrand: This is necessaryto test for situations where people have managed to get local connectivity (such as through the USB OTG port's usb0 IP address) but haven't actually gotten online. It's much easier to understand the implications of "I can't ping Google" than "huh, the apt-get is failing for some reason".

Kdisimone:
Added the ping google.com back in.  And a little note about the screws/nuts on edison board

PR #690, opened 2017-03-07
Putting in Mac autotune instructions

danamlewis:


Kdisimone:
Not currently working for me...after doing the mac steps, I get the following error:

scottleibrand:
Yeah, you can't use those quick install scripts on a Mac. I'm not sure you need anything they install: I think that's what the Mac specific steps are for...

Kdisimone:
Yup, just stopped to think logically now instead of simply following the steps.  Mac users need to be told to Skip steps 2A and 2B.  After that they should be fine

Kdisimone:
Works for the rest of the instructions.  Great!

scottleibrand:
Added commit https://github.com/openaps/docs/pull/690/commits/23fdbe04909d8878df393c548230a81d7cbbfe0a

Refactor Phase C Linux VM vs. Mac instructions
Creates a Step 0 : Decide where to run Autotune

Breaks apart Step 1 into Step 1a: Create a Linux VM or Step 1b: Prep your Mac

Moves quick-packages install into Step 1a.  Removes apt-get install jq step, as that's part of  quick-packages now.

PR #697, opened 2017-03-09
Clarifying Steps to skip for Mac users

Kdisimone:


scottleibrand:
I agree with not having Mac users run those two steps, but why are you proposing to re-order the steps for non-Mac users?

Kdisimone:
Just because if the "skip 2A and 2B" were up with the mac part previously...i think many people may miss it.  And if it comes at beginning of 2A and 2B, they may likely skip C on accident too.  Mostly just about what the easiest layout to keep the most users from missing the "skip this step" part.

scottleibrand:
Ok, I'm gonna try a bit of a refactor to avoid that issue.  One moment...

Kdisimone:
I'm good with whatever makes people skip.

scottleibrand:
Updated https://github.com/openaps/docs/pull/690 with that.  If you're good with that approach, we can close this one.

danamlewis:
Consensus in Gitter/elsewhere that https://github.com/openaps/docs/pull/690 ties in these changes + the other changes needed for Mac, so closing this and merging 690, can open additional PRs to further tweak Mac instructions as needed. Thanks!

PR #710, opened 2017-03-12
A series of changes to remove explicit Pi references and otherwise make Edison/Explorer Board the recommended setup

danamlewis:
Everyone agrees Carelink is non-optimal (range issues; ACK=0 bytes constant errors) and as a community, we are moving away from supporting running on Carelink. This PR is a batch of edits to remove any reference to Carelink beyond warning about the limitations; however, it can of course still be used and set up by hand/manually, but it won't be a well-supported option moving forward. This will also deprecate obvious references to Pi. The documentation still exists but no longer will be highlighted at the high level walkthrough view. All old Pi setups will continue to work; new Pi setups can be done; however, we only recommend that for people who already have CC debuggers as otherwise it is both more expensive and more troublesome and prone to errors.


Kdisimone:
100% agree.  Thank you!

PR #717, opened 2017-03-15
Update offline-looping-and-monitoring.md

danimaniac:
These changes attempt to consolidate and reorder the first portion of the xDripAPS instructions. They are relatively minor changes, but I believe it ready more clearly. Also, when following the existing doc I consistently inserted my OpenAPS rig hostname (instead of my online Nightscout instance api_secret) where the instructions asked for <api_secret>. Maybe others are more attentive, but I think my revisions will make the doc more specific in this regard.


danamlewis:
LGTM but will wait for a second thumbs up from @colinlennon or someone who's been using xdripAPS before merging. Thanks!

PR #718, opened 2017-03-17
Update Understand-determine-basal.md

parenthetic:
* fixed some inconsistencies, but then went overboard on grammar and parallelism...

### DIFF
--- a/docs/docs/walkthrough/phase-3/Understand-determine-basal.md
+++ b/docs/docs/walkthrough/phase-3/Understand-determine-basal.md
@@ -1,4 +1,4 @@
-# Understanding the output of oref0-determine-basal
+# Understanding the oref0-determine-basal logic
 
 The key logic behind any oref0 implementation of OpenAPS lies in the oref0-determine-basal.js code, which is what takes all of the inputs you've collected and makes a temp basal recommendation you can then enact if appropriate.  As such, it is important to understand how determine-basal makes its decisions, and how to interpret its output, so you can decide for yourself whether the recommendations it is making are appropriate for your situation, or if further adjustments are required before closing the loop or letting it run unattended.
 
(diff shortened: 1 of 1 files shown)


docs/docs/walkthrough/phase-3/Understand-determine-basal.md
```diff
@@ -12,40 +12,44 @@ The determine-basal algorithm requires a number of inputs, which are passed in J
 {"carbs":0,"boluses":0}
 {"delta":-2,"glucose":110,"avgdelta":-2.5}
 {"duration":0,"rate":0,"temp":"absolute"}
-{"iob":0,"activity":0,"bolussnooze":0,"basaliob":0}
+{"iob":0,"activity":0,"bolussnooze":0,"basaliob":0,"netbasalinsulin":0,"hightempinsulin":0,"time":"2017-03-17T00:34:51.000Z"}
 {"carbs_hr":28,"max_iob":1,"dia":3,"type":"current","current_basal":1.1,"max_daily_basal":1.3,"max_basal":3,"max_bg":120,"min_bg":115,"carbratio":10,"sens":40}
 ```
 
 * meal.json = `{"carbs":0,"boluses":0}`
```
scottleibrand: This should probably be updated to include mealCOB now that everyone is using AMA instead of the old meal assist algorithm.

parenthetic: Yeah, there are certainly other changes to be made in this documentation. The example input and output given in the docs don't match up all that well with what I see the pump-loop reporting, for example.

I'm only on day three running, and am still reading through the determine-basal code to get a handle on the logic. My plan is to understand it incrementally, starting off with just running in low-glucose suspend mode, without any bells and whistles, so I won't likely get to AMA logic and input/output for a while.

I just wanted to fix some internal inconsistencies first.

PR #727, opened 2017-03-20
Windows Flash Sheet

Kdisimone:
First stab...needs some formatting help and a second set of eyes.


gregtcm:
A suggestion would be to use ninite.com to install both Putty and 7Zip (rather than winzip).  https://ninite.com/7zip-putty/
I use to use winzip, but 7zip is much easier to use and you don't have all the nag screens.
We could add more programs from ninite.com, but these are the only two I noticed we need for this project.

I might have more suggestions when I get my edison later this week.

jcorbett80:
For certain versions of windows you can right-click on the file and choose extract all
unzip

scottleibrand:
I believe all supported versions of Windows support Extract All..., so we should make that standard and not recommend third-party unzipping tools IMO.

jcorbett80:
I agree but katie said her win7 did not. she looked all over for it. That is why I commented the way I did

jcorbett80:
also dfu-util. does it come in a win64 version. The link only takes you to a 32bit version

scottleibrand:
Sounds like her computer got messed up by an existing third-party unzip utility then: https://www.sevenforums.com/general-discussion/112220-restoring-extract-all-context-menu-windows-7-a.html

jcorbett80:
I acussed her of setting it beside a mac. I guess same thing

gregtcm:
On my Win 7 machine it is not there either.  It could be that a 3rd party application removes it.  FWIW.  Also, I've found 7zip much more useful, but I understand sticking with the standard built in windows way.

jcorbett80:
in section 1-3 you previously said put the flash window (command prompt window) on the left. however when you tell us to start the flash you in the window on the right "In your flash window on the right (command prompt window), enter flashall.bat"

Kdisimone:
Well, I also included it because I've seen others on Gitter comment they didn't have an unzip-at-right-click option either.  I was using my father in law's computer....it's not loaded with much of anything (except for a virus protection software).  I think we should leave some unzip options in there for people who don't have the right-click option appearing.  I'd be totally fine with redoing the screenshots and instructions to use 7zip.  I have some more edisons coming tomorrow and can redo screenshots.

Kdisimone:
I'd be happy to make it (unzipping third-party program) more of a side note in the PR

jcorbett80:
I am good either way

James

gregtcm:
Give ninite.com a try.  It is a very slick and seamless way to install aps.  The link I shared earlier will just do the two (Putty & 7zip), but from the main site (ninite.com) you could choose a different selection of applications.  Let me know what you think.

Kdisimone:
I'll be back to this with some edits when the edisons arrive in a couple days.  I'll also try to see how easy it will be to write up instructions for restoring extract all in the context menu

Kdisimone:
Fixed the language to be primarily Extract All as the instructions.  Oddly, when I got to the same Windows computer as I'd used before...the extract all command was available this time.  Once I worked on the computer for a bit to grab more screenshots (was thinking I might want some for 7zip if there was ever a need)...after that...the extract all command disappeared again.  I could bring it back quite easily for the jubilinux folder by setting properties for the folder.  But, the dfutil was not happening.  I think we should leave a short note for the third-party solution, but just keep it as that.  I removed WInZip, changed it to the 7-zip, and took out the screenshots of win-zip.  

also fixed the left, right error I had.  thanks for catching that.

Kdisimone:
Thanks @jcorbett80 for the dfu-util link.  I revised the write up to directly link to the two files we are interested in...thus totally negating any need to unzip dfu-util at all.  Great suggestion.  And I also added the Properties change for the jubilinux.zip folder if the extract all is missing on that one (so no third party app should be needed now in any situation...fingers crossed).

jcorbett80:
👍

danamlewis:
::pokes head back in here:: @scottleibrand @jcorbett80 @gregtcm thumbs up for merging now? looks like everything raised was edited/resolved, but let me know if there's anything here that's a blocker for merging (or otherwise if it looks good to go). I'll otherwise plan to merge tomorrow if no one speaks up otherwise.

gregtcm:
@danamlewis We probably should update the jubilinux section.  Also, flashing with the new jubilinux doesn't have the <ctrl>-D requirement.  Other than that, it looks good.  I didn't do a full run through, but I can do that this weekend if you want me to.  Otherwise, I think it's good enough to publish and we can always fine tune it later.

danamlewis:
Thanks @gregtcm; I removed the control-D reference and updated the jubilinux link. This LGTM now, @scottleibrand will you do a final eyeball on all the edits as fresh eyes?

PR #742, opened 2017-03-29
Updating NS page

Kdisimone:
Adding a how-to for setup and explicit variable lines for OpenAPS.  

I'm not sure whether the formatting for the .. note:: will work (saw it in other rtd sites and looked helpful?), nor am i totally sure that the picture formatting will work (but I sure hope so, i prefer the slightly smaller pics and having them centered.


Kdisimone:
Changed to dev branch to test out the img tags and notes

PR #749, opened 2017-04-03
Update autotune.md

mar5477:
Add link to sick day autosens


danamlewis:
Ah - good catch, thanks @mar5477!

PR #762, opened 2017-04-08
fix sed command to be compatible with jubilinux 0.1.1

scottleibrand:


esialb:
looks good to me for jubilinux v0.1.1 and later

esialb:
Need to add `-r` to sed and then it should be good.

PR #766, opened 2017-04-12
add cd oref0 to pull/clone oref0-dev

mar5477:
cd oref0 to look for package.json when running npm run global-install


scottleibrand:
Fixed via #794

PR #791, opened 2017-04-27
Adding phone compatibility list and BTAutoTethering details of Bluetooth Section

tim2000s:
Added a table of phones with their compatibility and details of the BTAutoTethering App for Android.


danamlewis:
you're awesome @tim2000s - thanks for adding this in

tim2000s:
Hmm. The markup that works for git doesn't seem to have translated to the docs for the table. What is the publishing system in use and I'll fix the markup.

danamlewis:
@tim2000s thx for fixing this via #798!

PR #807, opened 2017-05-02
Add loop-app-for-local-BGs-from-iPhone to offline looping page

danamlewis:
Props to @thebookins for this great method of using iphone for local BGs to be sent to the Edison rig!


danamlewis:
Flags for @sarahspins to take a look at; and maybe @Kdisimone to make sure my language around how we're using a Loopkit/Loop app non-traditionally won't confuse anyone.

danamlewis:
Going to add this since no objections/comments in 9 days; PRs welcome if/when clarifications and edits needed

PR #812, opened 2017-05-04
I think this was incorrect.

rsilvers129:
Not consistant with /root/src/oref0/myopenaps that is used in other places. But if I am wrong, then I am wrong.


scottleibrand:
No, the runagain script is created in your myopenaps directory with your particular settings. It's not part of the core source code. Take a look at your rig: you should have one there.

PR #814, opened 2017-05-06
Update troubleshooting.md

Sgoldberg89:
I had to enter this line of code: 
wget https://github.com/EnhancedRadioDevices/subg_rfspy/releases/download/v0.8-explorer/spi1_alt2_EDISON_EXPLORER_US_STDLOC.hex
after I already did the erase step, but I don't know why. I entered it again here BUT I know it's sequence thing and probably doesn't need to be entered twice, so please feel free to erase the one that shouldn't be there.


scottleibrand:
#813 came in just as you were submitting this.  I also added relative paths that should address the need to do the ~/src/ccprog stuff: https://github.com/openaps/docs/commit/ff5cf28f71df477ee7eb3701ee32d08bb0b7b949

Can you give that a try and see if we missed anything?

PR #817, opened 2017-05-10
ITTT needs to be mentioned up front.

rsilvers129:
I didn't know how to control the system until I read the doc for days. ITTT should be mentioned at the start for context on reading the rest of the docs. If you accept this change, then the phases may need to be renamed in other parts of the docs.


danamlewis:
Adding/renaming phases is a bit more than changing this page - point taken about IFTTT/control options being more prominent up front, though, so will close this PR and add it to the list for the re-org that's ongoing.

PR #824, opened 2017-05-13
Update offline-looping-and-monitoring.md

msrcgm:
Clarifying that the edison URL is just http:// + ip address


msrcgm:
So what is the procedure for finding URL of an Edison?

scottleibrand:
The Edison itself doesn't have a URL, just an IP address or hostname. It looks like the xDripAPS service listens on port 5000, so you would append `:5000` to the IP address of the Edison, as shown in the example you're changing.

PR #837, opened 2017-05-23
no certificate check for https

Kdisimone:


scripts/openaps-bootstrap.sh
```diff
@@ -26,5 +26,6 @@ echo -e "\nAttempting to bring up wlan0:\n"
 ifdown wlan0; ifup wlan0
 echo -ne "\nWifi SSID: "; iwgetid -r
 sleep 5
-cd /tmp/; wget http://raw.githubusercontent.com/openaps/docs/dev/scripts/openaps-install.sh; bash ./openaps-install.sh
+// TODO check for options to fix the certificate activation error message for https
```
scottleibrand: In bash, the comment character is # rather than //

Kdisimone: fixed...thanks!

scottleibrand:
FWIW, I noticed it because GitHub's syntax coloring was treating the `for`'s as if they were special.

PR #857, opened 2017-06-02
Update setup-edison.md

garykidd:


danamlewis:
I believe this fix is already reflected from another PR

PR #870, opened 2017-06-07
Update oref0-setup.md

davidkeddydb:
Install sudo if not on Edison board for initial set up


Aaybob:
They should be directed to 1-4, hostname for edison

danamlewis:
Resolved via: https://github.com/openaps/docs/pull/872

PR #873, opened 2017-06-07
Include lm-sensors

Kdisimone:


scottleibrand:
Let's do https://github.com/openaps/docs/blob/master/scripts/quick-src.sh#L4 as well to keep them in sync.

PR #880, opened 2017-06-12
Added more detailed information.

rsilvers129:


scottleibrand:
I would agree that we don't want to be overly specific or prescriptive here.

PR #890, opened 2017-06-15
Create offline-looping-and-monitoring.md

logichammer:
Updated info regarding Pancreabble's set up...Talking to @applehat, it sounds like Pebble support via the setup script isn't currently working but he is looking into it.  I couldn't get it to work via my attempts, FYI.

Using the tips I added above, I was able to connect and send notify messages...Haven't gotten Urchin working yet but will add info when I do.  All good?


logichammer:
Mark has implemented these changes on his side so we don't need to include them in the docs here...

PR #891, opened 2017-06-15
UAM and NS info

applehat:


danamlewis:
heh - can you re-target to master? That should fix all the excess here.

applehat:
@danamlewis done. lol

PR #911, opened 2017-07-01
Add notes about optional items

gregtcm:


danamlewis:
@gregtcm - what kind of range difference are you getting with the external antenna?

PR #916, opened 2017-07-03
Update oref0-setup.md

Aaybob:
Many users don't know why they can't get back to non dev


scottleibrand:
This is incorrect: you can go back to oref0 master.  Are people confused by seeing openaps 0.2.0-dev or whatever that is, and thinking they need to go back to 0.1.5 (they don't).

PR #930, opened 2017-07-09
Dev branch updates

Kdisimone:
* new loop non-errors


Kdisimone:
@scottleibrand ah, about those x12 and smbs...I'd thought it was intentional to keep them off the preflight check since there was a (high?) likelihood that users may not be very diligent about updating their profile for basal adjustments etc.  On that same topic, is there anything preventing x12 users from autotune running?  I wouldn't think so.

Kdisimone:
I like the edit...thanks

PR #944, opened 2017-07-14
Update troubleshooting-loop.md

cameronrenwick:


scottleibrand:
I'm actually not sure if this is correct: see discussion at https://gitter.im/nightscout/intend-to-bolus?at=596902bd3230e14f3a5c077d

You might want to do some more experiments with the Target BG range values and see what effect they have in NS...

PR #946, opened 2017-07-15
Update update-your-rig.md

rodzilla55:
Fixing broken link.... the link I substituted for the broken link is the same as the "up next" link at the bottom of the page.


danamlewis:
Good catch (also inadvertently fixed via all the redirects I just did, but we'll want to update the links when we catch them like this in the docs just so it doesn't confuse people in the future).

PR #953, opened 2017-07-17
Update autotune.md

cameronrenwick:
suggestions for new Autotuners


danamlewis:
This is actually up in Phase A description - doesn't need to be there in Phase C. (TY though for the PR!)

PR #957, opened 2017-07-19
Update ifttt-integration.md

cdmccook:
Searching for "maker" no longer yields the correct action service. Must now search for "webhooks," and icon is now blue instead of green. With apologies, didn't know how to update screen shots.


danamlewis:
Thanks, @cdmccook! maybe @Kdisimone will add to her screenshot wish list :D Otherwise, having the text content should be good enough in the meantime.

Kdisimone:
Hahah, that's the third time they changed the color!  Jerks!!

PR #981, opened 2017-07-26
fix broken images for token based authentications after doc update

PieterGit:
the image links were broken due to the big doc update


PieterGit:
@danamlewis can you merge this?

PR #990, opened 2017-07-31
Redirected urls

scottleibrand:
This update was done using:

`cat ../redirects.txt | while read line; do from=$(echo $line | awk -F \" '{print $2}'); to=$(echo $line | awk -F \" '{print $4}'); echo $from $to; grep -l $from -r * | grep -v bak | while read file; do echo From: $from To: $to in: $file; sed -i.bak "s^$from^$to^g" "$file"; done; done`


danamlewis:
Spot checked most of the links - they work

PR #992, opened 2017-08-01
remove spaces in URLs

scottleibrand:


scottleibrand:
If we want to do this, we'll also need to figure out how to update the redirects.

scottleibrand:
Should I just close this?

PR #1024, opened 2017-08-29
Update OpenAPS-install.md

danimaniac:
If you named your rig something other than the recommended "myopenaps" during oref0 installation, the shortcut commands must be edited to reflect your rig's name. Change "myopenaps" in the shortcut, do not change "openaps".

### DIFF
--- a/docs/docs/Build Your Rig/OpenAPS-install.md
+++ b/docs/docs/Build Your Rig/OpenAPS-install.md
@@ -226,7 +226,7 @@ These logs and other files are things you may frequently access. There are short
  edit-runagain => cd ~/myopenaps && nano oref0-runagain.sh
  cat-runagain => cd ~/myopenaps && cat oref0-runagain.sh
 ```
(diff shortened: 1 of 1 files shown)


scottleibrand:
We could even take directory naming out of the interactive setup script and only allow it to be changed via command-line flag.

danimaniac:
I named my multiple rigs consecutively...creativename1 and creativename2. Hence, the aliases return errors. Possibly a clarification regarding the naming of multiple rigs would be more appropriate? Maybe I'm not fully understanding the purpose of aliases - to save time and prevent typo's?

scottleibrand:
This has nothing to do with rig hostnames.  This is the name of the ~/myopenaps/ directory on each rig.  If you also named your myopenaps directory something custom, we recommend that you change it back to myopenaps the next time you run oref0-setup.

scottleibrand:
And yes, the aliases are to save typing for common administrative tasks that people need to do from the command line.

danimaniac:
Got it. I defer.

PR #1025, opened 2017-08-31
Adding link for HTTP Widget instructions

kcrcgm:
I found this github link that I think used to be in the docs but has disappeared. I'm adding it back into the section on how to get the HTTP Widget app to work for Android offline webpage monitoring.


danamlewis:
is HTTP widget different than HotButton? If so, can you actually add those instructions to this page vs. linking externally? Thanks!

kcrcgm:
Yes,  they are different apps, and I can add the instructions from this link into the http widget part of the docs. BUT I can't vouch for them working bc I haven't gotten it all to work perfectly.  Should I wait until I've tried to work out the kinks?

kcrcgm:
Ok, just ignore this pull request and I'll start over. Following the instructions in the link do not work,  which is I'm sure why it wasn't put in the docs. But it does have a nice screen shot of the app. I'll make a new pull request with that image and any other progress I make on this. Thanks!

PR #1030, opened 2017-09-01
Update glossary.md

jdhigh:
Alphabetize the glossary and add definition for COB; small wording changes/other minor grammar fixes

### DIFF
--- a/docs/docs/Resources/glossary.md
+++ b/docs/docs/Resources/glossary.md
@@ -1,23 +1,13 @@
 # Glossary
 
 
-## AP and OpenAPS high level terminology 
+## AP and OpenAPS high-level terminology 
 
 <b>APS or AP</b> - artificial pancreas system. A term for a closed-loop automated insulin delivery system in which temporary basal adjustments are used to maintain BG levels at a user-specified target range.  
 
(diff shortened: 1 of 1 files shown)


danimaniac:
I think an alphabetized glossary is a great idea. I also wonder if we should verify / update / revise some definitions at the same time? A true "Closed Loop", for example, does more than adjust basal rates but would also administer bolus insulin. It would do these tasks without requiring ANY user interaction...including carb input. Or should the content of the definitions be a separate PR? What do you think?

scottleibrand:
I would do a second PR for possibly-controversial definition changes.

danimaniac:
I'm good with that, Scott. Just wondered.

danamlewis:
@danimaniac by your definition, this IS a closed loop. It's actually a question of what kind of outcomes you're expecting - i.e. are spikes to 300+ ok? 150? Because it's really a behavioral choice about whether you input info, do anything or not. We want best outcomes, which is why we recommend behavioral input in a closed loop system. (Or did in the past ;) that might change w/ future versions). In any case, feel free to submit a separate PR if you really want to debate/dive into that/think it's important for the glossary. 

Merging and closing this PR. Thx @jdhigh

PR #1033, opened 2017-09-06
Understand iob calcs

parenthetic:
New page added to the docs in the "While You Wait For Gear" section.


scottleibrand:
Wow, nice work.  Just in time for throwing exponential curves into the mix with 0.6.0.  :-)

Do we want to add a placeholder note about the new exponential curves (and maybe a few links to issues and such) until someone has a chance to do a similar write-up for them?

parenthetic:
Thanks.
Not surprisingly, I don't know anything about the exponential curves development...
But adding a placeholder note for that development seems to make sense to me. Or are the curves I've documented/drawn getting tossed out completely in 0.6.0? In which case, it might not make sense to include this documentation at all.

The only issue I'm aware of about `iob` calculations is that `activity` and `iob` are based on independent functions. It seems to me that they should be explicitly linked to each other. I started tinkering with writing code to do just that, but haven't finished with it (yet!).

Are there explicit issues that folks have opened in the oref0 repo about the `iob` calculations that I'm not aware of? (I'm guessing so since you mention them, but I haven't looked -- sorry.) If so, linking to them also seems to make sense.

scottleibrand:
The exponential curves are intended to replace bilinear eventually, but we won't be disabling the bilinear ones any time soon. We won't be switching over the default right away, and don't yet have a plan for deprecating bilinear curves, but probably will do so at least for new users once we get some widespread testing of the new exponential ones.

parenthetic:
That sounds good. I'd be happy to replicate the code with the exponential curves, redraw the graphs, and update the docs when those come on-line. I took a peak in the oref0/dev branch, but didn't see any changes in the functional forms for the `iob` calcs. Could you point me to where those might be so I can get a look-see?  Thanks!

scottleibrand:
Exponential curves was merged into 0.6.0-dev via https://github.com/openaps/oref0/pull/568.  The PR to dev (https://github.com/openaps/oref0/pull/592) is waiting on the 0.5.3 release.

parenthetic:
Never mind. Found it. (oref0/scott-dev, right?)

parenthetic:
Whoops, didn't see your response before sending mine last message.

parenthetic:
Added paragraph explaining oref0, version 0.6.0, exponential insulin action curves, and pointed to oref0 issue #544.

parenthetic:
Hi Scott, 

I feel like your edits might make the graphs a little confusing. Most readers of the documentation are unlikely to know about the total.js code. (I didn't, for example.) And although total.js calculates variables called `activity` and `iob`, they are not the same concepts that I plot in the graphs.

The `activity` and `iob` concepts that I plot are both expressed in percentage terms. `activityContrib` and `iobContrib` convert those percentages into actual *units of insulin per minute* by multiplying by `treatment.insulin`. The variables in total.js are also in the same units: *units of insulin per minute*, but just a summation of multiple treatments, if there are/were multiple treatments.

(All of this, I'm sure you know. It just seems potentially confusing to me to define the units for the `activity` variable I'm plotting as *units of insulin per minute* and to then have the y-axis of the graph be "Percent of Insulin Used Each Minute.")

Perhaps another note at the end of the documentation pointing out that total.js exists and creates variables with the same names, but represent different things would be a better way to address this, and hopefully reduce any confusion for a newish reader of the documentation?

What do you think?

scottleibrand:
Ah, ok. I didn't realize that's what you were going for. Yes, I like that revised approach. Feel free to go ahead with that, or I might be able to do so by tomorrow.

parenthetic:
I might be able to take a first crack at the note tonight, but am less sure about how to unwind your edits.

scottleibrand:
Should be able to just `git pull` the branch from your GitHub repo and edit on top of them.

parenthetic:
Wish I saw your comment before I started working this evening!!
(Subversion is SO much simpler!)

I think I've got everything making good sense, but certainly check my work.

Thanks!

PR #1035, opened 2017-09-12
Update update-your-rig.md

jdhigh:
Is this helpful for some people? Should we update this page to include "Afterward, you also need to turn on the relevant settings in preferences.json." as a reminder for those who have setup and wish to continue using  SMB and UAM?


danamlewis:
Tweaked it to be a more generic "set your preferences" - great suggestion, thanks!

PR #1042, opened 2017-09-19
Fix for API_SECRET VM error

bdr1177:
Describes a fix for the following error when running autotune from a VM:  `ERROR: API_SECRET is not set when calling oref0-autotune.sh`


AllyCat09:
Hello I'm trying to find a solution to this same problem using Terminal on a MAC.  Thanks.

bdr1177:
@danielharrelson Thanks for catching that typo.  I went in and fixed if you.  If you have a chance, please look again and make sure it looks good to you know.

PR #1050, opened 2017-09-25
Update mac-flash.md

jdhigh:
Suggesting update in Mac flash steps to get Jubilinux 0.2.0. instead of 0.1.1.

### DIFF
--- a/docs/docs/Resources/Edison-Flashing/mac-flash.md
+++ b/docs/docs/Resources/Edison-Flashing/mac-flash.md
@@ -46,7 +46,7 @@ Building the software into your rig is comprised of three steps:
 
 The Edison comes with an operating system that doesn’t work easily with OpenAPS.  The first step is to replace the operating system with a new one.  This is called “flashing” the Edison.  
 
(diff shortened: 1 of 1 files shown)


scottleibrand:
Why?  There is nothing I know of in jubilinux 0.2.0 that improves things for OpenAPS users: it just required workarounds in the OpenAPS code to unbreak things.

danamlewis:
Per ITB discussions, since the fix is now in master (0.5.3), no reason not to just remove the warning of needing a particular version (which causes confusion). @jdhigh, can you tweak to remove the version specification, and just say download Jubilinux? Then we can get these in. Thanks for being willing to make the PRs!

jdhigh:
@danamlewis Ok, I'm still kinda new at this PR thing, but I think I correctly committed a change to just indicate download Jubilinux. HOWEVER, this COULD be problematic if Jubilinux creates a newer release than 0.2.0. I think that sometimes specificity is best to avoid future problems. (Which goes to Scott's question.) And some people will always opt for the most recent version, absent specific instruction, because the perception is that newer is better.

PR #1052, opened 2017-09-25
Update PC-flash.md

jdhigh:
Suggesting update in PC flash steps to get Jubilinux 0.2.0. instead of 0.1.1.


scottleibrand:
Why?  There is nothing I know of in jubilinux 0.2.0 that improves things for OpenAPS users: it just required workarounds in the OpenAPS code to unbreak things.

danamlewis:
Per ITB discussions, since the fix is now in master (0.5.3), no reason not to just remove the warning of needing a particular version (which causes confusion). @jdhigh, can you tweak to remove the version specification, and just say download Jubilinux? Then we can get these in. Thanks for being willing to make the PRs!

PR #1054, opened 2017-09-28
Steps have changed.

sdneufer:
ifttt_maker-webhook
An image needs to be changed and I have the image however I can't upload it to the images folder.  Deleted the connect clicks which are between a couple steps as they seem to be removed from the process.  Attached the new image that should replace IFTTT maker


Kdisimone:
@sdneufer Your PR currently only deletes the text...you instead should leave the text in place as it was and just replace the image.  If you upload an image of the same name as the original file was named, it will be replaced.

Kdisimone:
If you navigate to your patch-3 branch, then navigate to the iamges folder, you can use the upload file button to upload your new image file named the same as the existing file.  So, if you have new image files of IFTTT_connect1.png and IFTTTT_connect2.png....then you need to upload them there.

Kdisimone:
screen shot 2017-09-27 at 7 00 40 pm

Kdisimone:
Looks like you've deleted the image paths for two images though.  If you just want to replace them, edit the page to add the file paths back and just upload the images. 👍

sdneufer:
I was deleting the two images and text that did not appear when I made the new applet.
I uploaded the image that changed (the text was already changed... appears so anyway) so I did not change that text.

Kdisimone:
@sdneufer Was this the very first IFTTT applet you were creating? Those two steps you edited out only happen for the very first time you create an applet...so if you'd ever created one before with that account, those would've already been connected services so the link wouldn't have appeared.  Just double checking.

Kdisimone:
(or at least only used to happen for the very first applet...i haven't set up a new IFTTT account in a couple months to verify...hence my question. 😄 )

sdneufer:
I thought about that.  ok... checked it, created a new account...  back to the drawing board. ie... I will comment that the connect buttons only show up when creating the first applet in a new account.
Do I close this one and start over?
Should I choose a different branch?

Kdisimone:
Totally up to you....You can use your branch of patch-3 still. Any edits you make on that branch will be incorporated into this PR still. So you could edit back in those lines and add a note that those will only appear on the first applet.

Kdisimone:
Oops, sorry I was late in seeing this 😜

PR #1060, opened 2017-09-29
Putting image instead of chart since won't render in RTD

danamlewis:


danamlewis:
@parenthetic just FYI your beautiful chart works in markdown, but not ReadTheDocs. So, I [put the raw file of the chart as a file in this folder so we can update if needed](https://github.com/openaps/docs/pull/1060/commits/72619d15914ab1edb5ffa96c0d221535ee92c4ef), and just stuck an image in that section so people could see it.

parenthetic:
Bummer. Good catch and fix.

PR #1078, opened 2017-10-18
Update PC-flash.md

Aaybob:
Manual steps no longer used.


danamlewis:
Manual steps are still used by some people; leaving as is but adding the link. Thanks.

PR #1087, opened 2017-10-24
Use new software-properties-common package

cascer1:
python-software-properties has been replaced


cascer1:
I haven't checked installing with the package removed entirely, but it does fail completely without it because apt-get can't find the package and kicks the bucket

scottleibrand:
Ok.  I'd have a slight preference for removing it if it's no longer needed, if you want to check that.  I probably won't want to go reimage a rig just to check that, though, so I'm good with your solution as-is if you'd rather leave it.

cascer1:
I'll check if it works without it entirely later, my current setup will probably be reset soon anyway.

jbwittmer:
I manually installed the new package and got a rig working from scratch but didn't try going without it entirely.  So I can confirm replacing python-software-properties with software-properties-common works but I don't know about leaving it out entirely.

PR #1095, opened 2017-10-29
Added infor about RN4X and MIUI

vebaba:
Please, decide whether Note for MIUI users should be below table or as last Subtitle/article (below **Additional App requirement on Android to enable automatic BT Tethering reconnects**). Thanks!


danamlewis:
LGTM, thanks!

PR #1103, opened 2017-11-05
Update ifttt-integration.md

JoseeT1Bionique:
Yesterday when looking to do Buttons for my bolused carb counts, I could not find the information, I had to ask in Looped. When searching the docs, I did not make logical connection to Low Treatment. I get it now, but really I'm not looking to treat a low, I want to add the carbs I boluses for. Even if same code, I think this will make it easier to find.


danamlewis:
Perfect - this is exactly the kind of PR that I love! Any things you see like that where there are easy fixes to make things more clear! Thank you!

JoseeT1Bionique:
Thanks for the feedback, I will keep at it :)

PR #1124, opened 2017-11-21
more description for max basal rate

thisistw:
added bit more description to the example, but mainly added key search terms to the text to help numpties like me find the answer quicker.
Sorry - read the docs 101


thisistw:
That better

PR #1125, opened 2017-11-23
Added introductory Diabetes Math Section

Ebgineer:
New text and visuals to give background for non-technical users or new diabetics.  Intended as a preface to understanding-the-determine-basal-logic documentation.  Images were not loading in preview, so two references are present for each graphic.  The first is for debugging and will need to be removed from the final version.  The second is a relative reference that should work after merging, but does not work in the forked version.


danamlewis:
Closing; content is incorporated on other pages

PR #1140, opened 2017-11-29
Added note about running Autotune manually

jenm999:
Note about renaming autotune directory after running Autotune manually if you don't want OpenAPS to use the suggested settings.


scottleibrand:
I don't think that is correct.  It's the files in ~/myopenaps/settings that actually matter.

jenm999:
I was given bad advice then. :) Please correct!

On Wed, Nov 29, 2017 at 3:55 PM, Scott Leibrand <notifications@github.com<mailto:notifications@github.com>> wrote:

I don't think that is correct. It's the files in ~/myopenaps/settings that actually matter.

PR #1144, opened 2017-11-30
Reduce confusion about compatible pumps

endafarrell:
I have never before been a Medtronic pump user (always Roche, perhaps that's a European thing). 

When trying to find a pump suitable for use with OpenAPS I was confused about the "How to check pump firmware (check for absence of PC Connect)" section - I had asked the seller of a 722 pump about the "Connect Devices" menu under "Utilities". We both ended up confused.

The intention here is to clarify that the x12, x15 and x22 pumps do not need to have their firmware checked.

### DIFF
--- a/docs/docs/Gear Up/pump.md
+++ b/docs/docs/Gear Up/pump.md
@@ -16,7 +16,7 @@ NOTE: For European/WorldWide users who have access to a `DANA*R` insulin pump, y
 
 ## How to check 523/723 & 554/754 pump firmware (check for absence of PC Connect)
 
-Note: his does not apply to the 512/712, 515/715 nor 523/723 pumps as their menues do not have a "Connect Devices" sub-menu within "Utilities". Any functioning x12, x15 or x22 pump will work, there is no need to check the firmware.
+Note: his does not apply to the 512/712, 515/715 nor 522/722 pumps as their menues do not have a "Connect Devices" sub-menu within "Utilities". Any functioning x12, x15 or x22 pump will work, there is no need to check the firmware.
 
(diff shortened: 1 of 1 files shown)


docs/docs/Gear Up/pump.md
```diff
@@ -14,7 +14,9 @@ Currently, only the following Medtronic MiniMed models allow us to remotely set
 
 NOTE: For European/WorldWide users who have access to a `DANA*R` insulin pump, you may be able to use AndroidAPS, which leverages OpenAPS's oref0 algorithm but allows you to interface using an Android phone and Bluetooth to communicate directly with the `DANA*R` pump. [See here for instructions and details related to AndroidAPS](https://github.com/MilosKozak/AndroidAPS).
 
-## How to check pump firmware (check for absence of PC Connect)
+## How to check 523/723 & 554/754 pump firmware (check for absence of PC Connect)
+
+Note: his does not apply to the 512/712, 515/715 nor 523/723 pumps as their menues do not have a "Connect Devices" sub-menu within "Utilities". Any functioning x12, x15 or x22 pump will work, there is no need to check the firmware.
```
scottleibrand: This should say 522/722.

endafarrell: I'm confused. All of the 522/722 firmwares are good, so my understanding is that there is no need to check the firmware on these. 

Only the 523/723 and the 554/754 models have to have the older firmwares - hence only those models need to be checked. 

Have I missed something?

scottleibrand: Your logic is correct. You just typo’d (“nor 523/723”).

danamlewis:
This change actually will cause more confusion. Many people do need to check PC connect presence in their pumps, since some sellers will mis-name pump models - so leaving this as a generic "how to check" will continue to be helpful. Closing. Please reopen and adjust if you'd prefer to change the content to clarify that the *absence* of the connect devices menu is a good thing, if that is currently unclear.

endafarrell:
ACK - #1148 instead.

PR #1148, opened 2017-11-30
Reduce confusion about compatible pumps, yet keep importance of verif…

endafarrell:
…ication

Based on #1144 comments by Dana, keeping the importance of verification, yet reflowing to provide clarity that - once you're certain that the pump is indeed one of  512/712, the 515/715 and 522/722 (I had pictures of mine before purchasing), then you do not need to ask follow-up questions about settings in a menu that doesn't exist (which caused confusion for both of us - these docs hadn't previously mentioned that the "Connect Devices" menu does not exist on the 512/712, the 515/715 and 522/722 models.


danamlewis:
Great, thank you

PR #1130, opened 2017-11-26
install oref0 0.6.0

scottleibrand:

PR #1150, opened 2017-12-05
Autotune Docs oref0 version change

musicman0329:
Updated the version of oref0 in autotune docs with version 0.6 instead of versions 0.4 and 0.5


danamlewis:
(Closing, per #1130 )

PR #1153, opened 2017-12-06
Lessons from a recent (Dec 2017) build

endafarrell:
Clearer cable instructions, newer links to the latest jubilinux files.


endafarrell:
:-)

PR #1159, opened 2017-12-08
Update how-openaps-works-overview.md

GlennCoco:
Links to various parts of the instructions to make it a little easier to navigate, I also would like to put together a section under build your rig with construction instructions, pictures and common questions asked my the community, I think this is important, once complete I will add a hyperlink in this page for 'physically put your rig together' to that page.
Please let me know if this is something already in the works to avoid double up.

### DIFF
--- a/docs/docs/Understanding OpenAPS-Overview/how-openaps-works-overview.md
+++ b/docs/docs/Understanding OpenAPS-Overview/how-openaps-works-overview.md
@@ -3,9 +3,9 @@
 How do you make decisions about your diabetes? You gather data, crunch the numbers, and take action. 
 
 A DIY loop is no different. It gathers data from:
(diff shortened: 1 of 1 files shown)


danamlewis:
There is some info on http://openaps.readthedocs.io/en/latest/docs/Resources/switching-between-DIY-systems.html and http://openaps.readthedocs.io/en/latest/docs/Resources/Edison-Flashing/mac-flash.html for pictures/putting rigs together - I'd be fine with adding a page with the info from those (and anything else you want to add) and putting it in the top section and linking it as described.

GlennCoco:
Great sounds good, I will get started on a draft this week. As there is no existing page for it, what format and where would you like me to put it together?

PR #1173, opened 2017-12-15
Update on-the-go-wifi-adding.md

increddibelly:
I've spent hours trying to connect. wpa_passphrase fixed it. Also, how do I PROPERLY add an image?


increddibelly:
my wifi only started working because I sat down right next to the router. It has nothing to do with WPA-PASSPHRASE.

PR #1179, opened 2017-12-20
Update PC-flash.md

increddibelly:
took me long enough, but at least I'm online. I sincerly do not like Linux. In fact, I feel a lot stronger, but children may be reading this.

### DIFF
--- a/docs/docs/Resources/Edison-Flashing/PC-flash.md
+++ b/docs/docs/Resources/Edison-Flashing/PC-flash.md
@@ -184,15 +184,16 @@ network={
 }
 ```
 
-most recent installations require a wpa passphrase instead of a wpa pre shared key.
+you can also use a wpa passphrase instead of a wpa pre shared key.
 from a root prompt, type
 ```wpa_passphrase ssid password```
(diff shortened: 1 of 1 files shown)


docs/docs/Resources/Edison-Flashing/PC-flash.md
```diff
@@ -184,6 +184,33 @@ network={
 }
 ```
 
+most recent installations require a wpa passphrase instead of a wpa pre shared key.
```
scottleibrand: I don't think anything has changed here.  both configuration methods work (except of course if you have funny characters that are hard to represent in plain text)

increddibelly: I agree. I thought I had my finger on it. I find it surprisingly complicated to get wifi up and I've been a professional nerd all my life. Color me surprised. I should probably rephrase the useful bits of my pull request.

increddibelly:
my wifi only started working because I sat my LPOF down right next to the router. none of my suggestions in these commits are the actual fix. If any of the info seems useful, please edit or merge.

PR #1185, opened 2017-12-27
Bolus wizard caution

jpcunningh:
Added a caution in the docs where it discusses entering carbs via the pump's bolus wizard to ensure that only the carbs and entered into the bolus wizard and not the current BG. This caution is intended to prevent someone from double dosing, once by OpenAPS temp basals and once by the pump Bolus Wizard to bring a higher BG to a lower target.


jpcunningh:
See April 28, 2017 and May 26, 2017 discussions on https://gitter.im/nightscout/intend-to-bolus. We are on our 4th day and just experienced this. Realizing what happened, I was able to find the discussion, but couldn't find anything clear in the docs.

jpcunningh:
You're right. That's another good place to put a caution. I'll add it and push it to this pull request. Thanks!

docs/docs/While You Wait For Gear/preferences-and-safety-settings.md
```diff
@@ -194,7 +194,9 @@ This feature was previously enabled by default but will now default to false (wi
 
 #### override_high_target_with_low: 
 
-Defaults to false, but can be turned on if you have a situation where you want someone (a school caregiver, for example) to use the bolus wizard for meal boluses. If set to “True”, then the bolus wizard will calculate boluses with the high end of the BG target, but OpenAPS will target the low end of that range. So if you  have a target range of 100-120; and set this to true; bolus wizard will adjust to 120 and the loop will target 100. If you have this on, you probably also want a wide range target, rather than a narrow (i.e. 100-100) target.
+Defaults to false, but can be turned on if you have a situation where you want someone (a school caregiver, for example) to use the bolus wizard for meal boluses. If set to “True”, then the bolus wizard will calculate boluses with the high end of the BG target, but OpenAPS will target the low end of that range. So if you  have a target range of 100-120; and set this to true; bolus wizard will adjust to 120 and the loop will target 100. If you have this on, you probably also want a wide target range with a high end such as 250, rather than a narrow (i.e. 100-100) target.
```
scottleibrand: This default behavior was updated without changing the documentation, such that it now defaults to true. We should update this paragraph as well.

jpcunningh:
Will do!

jpcunningh:
I looked through the code to find override_high_target_with_low so I could describe the alternate behavior. It appears that maybe that preference is no longer used and wide_bg_target_range is used instead? It seems to accomplish the same thing but opposite logic?

Am I missing something?

scottleibrand:
Oh yeah, that is correct.

jpcunningh:
I delete the paragraph on override_high_target_with_low and moved the warning to wide_bg_target_range.

I also reworked the wording a bit to be a little easier to follow. I'm going to update the other files with the new wording as well.

PR #1189, opened 2017-12-29
Update all-computers-flash.md

increddibelly:
I finally found the source of my wifi troubles : EDI2.LPOF has about a foot of WiFi range. 
I'd love to apply a foot to the backside of the persons who decided that some IOT devices do not need WiFi antennas.


danamlewis:
We've got notes on the Edison gear page about the types of Edison's recommended.

PR #1195, opened 2018-01-07
small change to wifi connection script

ZandMarc:
wrong directory mentioned


ZandMarc:
Thanks! I've corrected it to the full path (with /root)

PR #1196, opened 2018-01-10
Update all-computers-flash.md

kellyisworking:
Fix link to https://openaps.readthedocs.io/en/latest/docs/Resources/Edison-Flashing/mac-flash.html


scottleibrand:
A possibly better way to do this is to replace the html with md as I just did in dev in https://github.com/openaps/docs/commit/9934e0aa

PR #1209, opened 2018-01-28
Updated xDripAPS documentation

philipgo:
The xDripAPS documentation was outdated and caused confusion since OpenAPS does no longer forward CGM data received via xDripAPS to Nightscout. Also removed references to outdated versions of the XDrip app and clarified that this app differs from the iOSxdripreader app.

### DIFF
--- a/docs/docs/Customize-Iterate/offline-looping-and-monitoring.md
+++ b/docs/docs/Customize-Iterate/offline-looping-and-monitoring.md
@@ -134,23 +134,20 @@ If you wish to set a true temporary target while offline, you can do so by ssh'i
 
 ### xDripAPS for offline BGs and looping for Android users
 
-**Note as of 1/26/17:** The below documentation is WIP and needs additional testing.
+Do you use OpenAPS and the xDrip+ Android app? By deafult, the xDrip+ Android app uploads CGM data to an online Nightscout instance, OpenAPS then downloads this data for use in your online loop. 
 
-Do you use OpenAPS and the xDrip/xDrip+ Android App? Until now this required an internet connection to upload your xDrip/xDrip+ Android App CGM data to an online Nightscout instance (the OpenAPS community recommends utilizing Heroku). Then your data was downloaded to your OpenAPS rig for use in your online loop. The xDripAPS code resides on your OpenAPS rig and allows the direct transfer of xDrip/xDrip+ Android App CGM data to your OpenAPS rig without an internet connection. xDripAPS creates an offline OpenAPS rig which utilizes a "local" or "personal" network (WiFi hotspot or Bluetooth PAN tethering) for direct communication between the xDrip/xDrip+ Android device and the OpenAPS rig. Data which is 'missing' from Nightscout will be uploaded when the OpenAPS rig regains internet connectivity.
-
-The OpenAPS community recommends an Explorer Board / Intel Edison rig, but xDripAPS also works with a Raspberry Pi rig.
-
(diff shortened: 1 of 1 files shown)


musicman0329:
Oops. I just submitted a PR with many similar changes regarding xDripAPS. Missed this one, sorry. I'll delete mine...I like yours better. ;) 

@danamlewis If this is all set, can we have this merged sooner rather than later? There has been several users in the Looped FB group asking for updated documentation on how to set this up. https://www.facebook.com/groups/TheLoopedGroup/permalink/1996718587211464/

drnoname82:
@musicman0329 Sorry for the time you spent!
@danamlewis Is there anything that I should change before we can get this merged?

@musicman0329 Are you sure about entering carbs via XDrip+? It did not work for us and I was under the impression that xDripAPS generally does not accept treatments. Also, soon OpenAPS may not upload xDripAPS data to Nightscout anymore ( https://github.com/openaps/oref0/pull/938 ), so I think we should maybe test this method again after the changes to the upload method have been implemented.

musicman0329:
@danamlewis commenting as requested...I can confirm that this PR fixes several inaccuracies in the docs (mostly related to xDrip+ since some labeling/locations of things have changed), as well as including some newer instructions that weren't present before (xDrip+ vs xDrip...multiple URLs in xDrip+....etc). Thank you!

philipgo:
@musicman0329 Line 183 was „Use both API Upload (REST) and MongoDB“, is that what you refer to? OpenAPS does not upload values received from xDripAPS anymore, they must be uploaded by the xDrip+ app so the method described is unnecessary. Or are you referring to something else?

docs/docs/Customize-Iterate/offline-looping-and-monitoring.md
```diff
@@ -180,21 +177,25 @@ A few notes to clarify:
 
 ![REST API Upload setting](https://github.com/colinlennon/xDripAPS/blob/master/xDrip_REST_API_cropped.png "REST API Upload setting")
 
-If using xDrip+ navigate to Settings > Cloud Upload > MongoDB and uncheck the "Skip LAN uploads" option. Do not turn on the "Enable Nightscout Mongo DB sync" option. Next, navigate to Settings > Cloud Upload > API Upload (REST) and uncheck the "Skip LAN uploads" option. NOTE: if you don't have these options, update to a recent version of the xDrip+ app. These options were added to a nightly build in December 2016.
```
musicman0329: @philipgo this is the line that I'm referring to

philipgo: Using Rest-API and Mongo uploads simultaneously does not make sense anymore because OpenAPS does not upload xDripAPS values anymore, so letting the xDrip+ app upload to two Rest-API destinations is the method of choice. MongoDB upload is discouraged by the Nightscout developers. 

I dropped explaining the „Skip LAN uploads“ section because I thought that it was unchecked by default, but please correct if I am wrong.

musicman0329: Correct, using both Rest-API and MongoDB is not recommended. However, the Skip LAN uploads option is actually CHECKED by default. So you or I need to re-add the second half of that section that refers "Skip LAN uploads" for Rest-API, or xdripAPS won't work.

philipgo: Could you double check that, maybe there has been a change? The full option is "Skip LAN uploads...when there is no local network connectivity", so as long as there is a local network connection, they should not be skipped when the option is checked, unchecking the option should only cause xDrip+ to attempt to upload them even if there is no local network connection. We have been using xDripAPS for a long time and I am pretty sure we had this option checked, but I cannot verify this right now.

musicman0329: @philipgo I tested it with the option both checked and unchecked, and it worked both ways. Good catch, thanks!

philipgo: Cool, thanks for testing!

PR #1246, opened 2018-03-18
adding spoof warning to flash

hilarykoch:
Dana- please don't review yet, still working on images


hilarykoch:
Ok, good to go.  Thanks

PR #1258, opened 2018-04-12
Update Hardware.md

statuur:
As promised hereby a new infographic and a short description about read-only pumps. Hopefully you consider it an improvement. If there is anything als it can do please ask. The png is located at my own server. Don't know how to get it at the right place. My english is not flawless. There could be some typos in the text.

### DIFF
--- a/docs/docs/Gear Up/hardware.md
+++ b/docs/docs/Gear Up/hardware.md
@@ -13,11 +13,14 @@ If you come across something that doesn't seem to work, is no longer available, 
 
 To start, here is a high-level guide for understanding if your pump is compatible for OpenAPS. To be compatible, we must be able to send remote temporary basal rate commands to it.
 
-!["Can I do OpenAPS with this pump?"](../Images/Can_I_close_the_loop_with_this_pump_July_13_2017.jpg)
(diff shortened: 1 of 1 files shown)


danamlewis:
Thanks @statuur! I can't open the raw file (but thank you for sharing it), so I made some notes on top of the image file: [Dana edits_hardware.md.pdf](https://github.com/openaps/docs/files/1964864/Dana.edits_hardware.md.pdf)
 
Key comments:
- Fixed some typos and otherwise edited text that I had been wanting to edit!
- Can we add a 640 (same as 670) pump visual to the top as well?
- Let's make the 5/723 and 5/754 not have the color split, as it makes it look like 7-series is not compatible.
- I like the read-only section addition! Let's also add a green arrow there to flow that step to OpenAPS.org

I'm wondering if it would be too much to also add the DANA*R and DANA*RS and Combo pumps on the right, to point to AndroidAPS? Example poorly drawn :) below. 

screen shot 2018-05-01 at 11 53 46 am

Let me know if you have any questions about the edits I suggested! Thanks again for working on this! (Also feel free to PM me on Gitter if it's easier to discuss in real-time.)

PR #1271, opened 2018-05-02
Added additional IFTTT script for Afrezza

sjolundjohn:
For Afrezza or can also be used for insulin pen correction bolus.


sjolundjohn:
Following discussion with Dana and Scott - I was made aware that this action would not be put into the calculation for OpenAPS, rather it creates a note in Nightscout.

The suggested path to solving this was to put a temp target for 30-45 minutes of 200. This would allow OpenAPS sufficient time to see what is happening with the extra insulin IOB, while reduce dosing.

Scott also highlighted:
if you don't have profile.exercise_mode or profile.high_temptarget_raises_sensitivity set, then you might need a high temp target of 160 or more. if you do, then something like 140 would probably be sufficient

allowSMB_with_high_temptarget is also relevant: if that's not set, you won't get any SMBs as long as a >= 111 high target is set

PR #1272, opened 2018-05-05
Update hardware.md

statuur:


danamlewis:
Ooh, spotted one more thing that needs tweaking - it says 544 instead of 554 - can you change that? Thanks! @statuur

PR #1276, opened 2018-05-15
Carelink update

MilosKozak:


juehv:
As far as I know, the code just checks if the word "carelink" is part of the punp.ini. I think carelin = on sends the wron signal to the users, since they might guess carelink = off would deactivate it. I suggest to user "radio_type = carelink".

scottleibrand:
I would agree that would be slightly better language and have the same effect. Please do PR in a change.

danamlewis:
(Working on a code fix for this, help testing https://github.com/openaps/oref0/pull/1011 would be appreciated to see if this resolves automatically upon setup rather than requiring people to manually edit the file. Thanks!)

PR #1279, opened 2018-05-16
Change iOSxdripreader to Spike

philipgo:
Changed iOSxdripreader to Spike since iOSxdripreader is not maintained anymore.


musicman0329:
@philipgo I just made a similar fix as a part of a larger PR for that page. Is it alright to just add the link you provided for Spike so that I don't have to rework my whole PR? Also, if you have a minute to review mine and make sure it's all set it would be appreciated.

philipgo:
Of course, I will review yours in a few hours!

PR #1294, opened 2018-06-11
Updated xdrip-js, Lookout, and Logger locations

jpcunningh:
xdrip-js, Lookout, and Logger are now hosted in the xdrip-js organization.


danamlewis:
Thanks! This LGTM but keep wanting to ask "BG data is sent to both OpenAPS (via xDripAPS) locally," since we're discontinuing xdripAPS, I think, the references to xdripAPS need removing on this page, too? (@scottleibrand correct me if I'm wrong)

scottleibrand:
As I understand it, xdripaps will still exist, we just won't upload from the rig to NS (xDrip+ has to do that).

jpcunningh:
Lookout and Logger should both be good to go with the OpenAPS change. Both upload BG data directly to NS in addition to sending it to OpenAPS via xDripAPS.

Logger supports back filling NS when connectivity is down, and I'm currently testing a branch for Lookout to back fill NS.

@efidoman (Logger) and I (Lookout) are both running oref0-ns-loop instead of the old openaps ns-loop without any issues.

PR #1296, opened 2018-06-13
Added specific reasons for Edison over Pi HAT

Lytrix:
I will be starting with a build to use OpenAPS setup, but in the documentation I was searching for reasons to go for the Edison instead of the Pi HAT, and found with thanks to @scottleibrand on Gitter!

### DIFF
--- a/docs/docs/Gear Up/edison.md
+++ b/docs/docs/Gear Up/edison.md
@@ -2,8 +2,18 @@
 
 You have two main options for hardware:
 
-* For most of 2017, the preferred rig was an Edison + Explorer Board. This is still a highly recommmended rig, if you can find an Intel Edison. [See below for details on Edison setups](http://openaps.readthedocs.io/en/latest/docs/Gear%20Up/edison.html#hardware-information-for-intel-edison-based-setups).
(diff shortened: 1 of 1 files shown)


docs/docs/Gear Up/edison.md
```diff
@@ -14,7 +24,7 @@ Summary of what you need for a Pi/HAT rig:
 * SD card
 
 #### HAT: 
-As of April 2018, there is be a Pi+HAT rig as an option for closing the loop with OpenAPS. The HAT can be ordered from the same place that makes the Explorer Board ([click here to pre-order](https://enhanced-radio-devices.myshopify.com/products/900mhz-explorer-hat?variant=1950212653065)). We call it the "Explorer HAT", to differentiate from the Explorer "Board" that goes with the Edison (see below).
+As of April 2018, there will be a Pi+HAT rig as an option for closing the loop with OpenAPS. The HAT can be ordered from the same place that makes the Explorer Board ([click here to pre-order](https://enhanced-radio-devices.myshopify.com/products/900mhz-explorer-hat?variant=1950212653065)). We call it the "Explorer HAT", to differentiate from the Explorer "Board" that goes with the Edison (see below).
```
scottleibrand: The HATs are shipping, so "is" would be the right word here.

PR #1297, opened 2018-06-14
Update autotune.md

bionictom:
Added information how to use autotune inside nightscout. Added "detail" tags for increased readability


bionictom:
Understood, thanks for the explanation and feedback.

danamlewis:
(Closing as we'll be pointing people to AutotuneWeb as the easiest way to run autotune for non-OpenAPS'ers)

PR #1301, opened 2018-06-23
Remove the note about country code, since latest PiBakery no longer r…

zgohr:
…equires it for WiFi to start.


scottleibrand:
@kdisimone can you double check that this is no longer needed? I haven’t used PiBakery myself yet.

Kdisimone:
I use macOS and have the country code box. 

I can’t text this right now because my piHAT is out for repairs

zgohr:
Yes, OSX version 0.3.6

image

Also, as far as I could tell, there was no locale set (got a lot of warnings during install/setup that no locale was set) but everything seemed to have worked fine...

PR #1305, opened 2018-06-27
Update pi-install.md

Aaybob:
different order of prompts first install on windows, ssh just needs the one word hostname, mentioned warning message just so its not a surprise... this was windows experience

### DIFF
--- a/docs/docs/Build Your Rig/pi-install.md
+++ b/docs/docs/Build Your Rig/pi-install.md
@@ -46,7 +46,7 @@ Give the rig a couple minutes to boot up.  Once the green LED stops blinking as 
 
 On Mac, open Terminal and use `ssh pi@raspberrypi.local`
 
-On Windows, use PuTTY and establish an SSH connection, with username `pi`, to hostname `raspberrypi.local`. 
+On Windows, use PuTTY and establish an SSH connection, with username `pi`, to hostname `raspberrypi.local`. Respond YES to the warning about computer security.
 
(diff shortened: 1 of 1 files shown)


docs/docs/Build Your Rig/pi-install.md
```diff
@@ -46,7 +46,7 @@ Give the rig a couple minutes to boot up.  Once the green LED stops blinking as
 
 On Mac, open Terminal and use `ssh pi@raspberrypi.local`
 
-On Windows, use PuTTY and establish an SSH connection, with username `pi`, to hostname `raspberrypi.local`. 
+On Windows, use PuTTY and establish an SSH connection, with username `pi`, to hostname `raspberrypi.local`. Respond YES to the warning about computer security.
```
scottleibrand: I don't want to train people to blindly answer security questions.  (That's generally a good way to get your money stolen and your computer locked with ransomware.)  Can you capture exactly what it is that the warning says/asks, so we can note here what you're saying YES to?

musicman0329: It's a certificate error. It's a common occurrence with self-signed or unsigned certificates, but you're right, we need to specify exactly what warning will appear (as best as possible, because there will be differences between the keys for each device). The best way to reproduce it is to use a computer that has never ssh'ed into that particular rig before. That should trigger the warning message.

musicman0329: image

musicman0329: This is from my pi rig. @Aaybob can confirm whether this is the correct error message.

PR #1313, opened 2018-07-10
Creating xml folder for image/flowchart sources

danamlewis:


danamlewis:
Due to crickets, merging as is. If you'd like to improve the images - please grab the xml files and edit and upload replacement xml and image files :)

PR #1321, opened 2018-07-20
Added Pi0 Battery and SD Card recommendations

Ebgineer:
Info on battery, case and SD card options with recommendations for each.


danamlewis:
Yay! thank you @Ebgineer

PR #1323, opened 2018-07-24
Updated bluetooth tethering docs 

tim2000s:
Added a section on checking for update_config=1
Added new phone to the table
Added an extra app to the auto tethering apps section

Also run through the process as described to see if anything needed changing. Once again, it's run through with no issues for me starting from nothing being connected, so I suspect that issues people are seeing with tethering are linked to phones as much as they are linked to what the Edison and EXB are doing.


danamlewis:
nice. thanks @tim2000s

PR #1338, opened 2018-08-20
mmtune update

jpcunningh:
This PR is dependent on openaps/oref0 PR, https://github.com/openaps/oref0/pull/1085


PieterGit:
@jpcunningh docs should work for both 0.6.x and 0.7.x rigs, so you duplicate the commands

PR #1340, opened 2018-08-26
Fixed markdown formatting in Pump rig troubleshooting.

jpcunningh:
The markdown isn't as forgiving in readthedocs as it is in github.

This attempts to fix some of the numbering issues and command lines that extend pass the right side of the screen.


danamlewis:
Rebased to master

PR #1351, opened 2018-09-01
Added `sudo` and additional steps

yeagerm:
For cleaner installs of oref0 dev on the Mac


docs/docs/Customize-Iterate/autotune.md
```diff
@@ -194,7 +194,17 @@ npm list -g oref0 | egrep oref0@0.5.[5-9] || (echo Installing latest oref0 packa
 
 ```
 cd ~/src && git clone git://github.com/openaps/oref0.git || (cd oref0 && git checkout dev && git pull)
-cd ~/src/oref0 && npm run global-install
+cd ~/src/oref0 && sudo npm run global-install
```
scottleibrand: I don't believe this is correct.  The `npm run global-install` is an npm alias for some things that should be sudo'd, and others that shouldn't.

PR #1361, opened 2018-09-19
Re-Run Autotune After Initial 'One-Off' Set-up

GitDRW:
This adds a step by step guide to how to re-run Autotune after initial 'One-Off' set-up.  I've put together a simple step-by-step guide on how to re-run Autotune manually once it has been set-up for 'One Off' use according to the OpenAPS Docs - these steps are for non-OpenAPS users to help tune settings for Loop or non-loopers. These steps are probably really obvious for the 'techies' out there but I'm not one of those and it took me a bit of figuring out - I thought this might help someone else after I put it together for my use. Sorry about my lack of formatting! It looked okay in the Word document version I put together but all the formatting vanished when pasting it here. As I mentioned, tech isn't my strength so I don't know how to format it nicely here to match the rest of the Docs. You're welcome to incorporate it if you like, I won't be the least bit offended if you don't or if you revise it significantly. I do think some instruction on how to do this and indeed pointing out that it can be done without repeating the entire initial One-Off Autotune set-up process might encourage more Loop users and 'non-loopers' to use Autotune. I was pleased to figure out I didn't have to type all the pump basal settings command lines etc each time I wanted to run Autotune manually, that in fact I could just edit the numbers then run it again.


GitDRW:
Thanks Dana, I'm glad that can be of use to the community. Since putting it together I actually now rarely use Autotune via that method as Mark Carrington has put together a very easy to use online version which you're probably familiar with. A link to his web-based version of Autotune in this section of the Docs would make it even easier for people. Here is the link if you want to check it out - https://autotuneweb.azurewebsites.net/ . I have no idea if there are issues with copyright if the link is shared in the docs. It is wonderfully easy to use that's for sure. Thanks as always Dana, Denton.

PR #1363, opened 2018-09-24
Add warning  for Enlite users

juehv:
OpenAPS is not working with underclocked Pi rigs since the loop runs so slow that bg values get outdated till they are processed. This issue has been reported several times on gitter and observed by myself.


juehv:
@scottleibrand what does "changes approved" without a merge mean ?

scottleibrand:
That it looks good to me, but I usually let @danamlewis review and merge docs PRs.

juehv:
Ok, thanks for the info 👍

PR #1364, opened 2018-09-27
Revert 1345 more monitoring options

danamlewis:


scottleibrand:
Why do we want to revert this in dev?

PR #1366, opened 2018-09-29
Autotune tune insulin description

jpcunningh:


jpcunningh:
@scottleibrand, please check the additional words to describe how useCustomPeakTime affects autotune now.

jpcunningh:
With that in mind, does it make sense to add some explanatory comment about
how it pulls in the dia and peak from the pumpprofile unless
--tune-insulin-curve is used or useCustomPeak is set?

scottleibrand:
Without useCustomPeakTime, DIA comes from the pump, but peakTime comes from the defaults for the curve selected in preferences.json.

jpcunningh:
How does that look?

PR #1373, opened 2018-10-07
Discourage unnecessary use of --categorize-uam-as-basal=true

scottleibrand:
Many people seem to be confused as to whether they should use --categorize-uam-as-basal=true or not.  This PR clarifies that --categorize-uam-as-basal=true is only needed in rare cases where the majority of basal data is being excluded as UAM.


Illiteratus:
Great ! Well done !!   Thank you....
Now it's much easier to understand the purpose of the command. 
Additionally, due to the psychological fact that most people think they do "everything" right (including the estimation of their basal rates), they will not mess around with that command if you describe it in a way that it has to be used if they don't have a clue about what's going on with their body.
They will probably use it just if they're really desperate and admit they messed up......
Thanks again for clarifying.....

scottleibrand:
Yes, I think "of course I do everything right" was leading some people to use --categorize-uam-as-basal=true before (based on "yes, I enter all my carbs").  Hopefully this now will lead those same people to leave it at --categorize-uam-as-basal=false (based on "yes, my basals are pretty good").  :-)

MilosKozak:
common use should be then:
if you announce all carbs and have uam off?

scottleibrand:
No, the use of the UAM feature for dosing is completely orthogonal to the exclusion of unexplained rises in autotune. Neither is really primarily about completely unannounced meals, and they do things differently.

PR #1378, opened 2018-10-26
Update quick-packages.sh

sethgagnon:
script will detect if running on WSL and use nodejs package instead of node-legacy package

### DIFF
--- a/scripts/quick-packages.sh
+++ b/scripts/quick-packages.sh
@@ -1,15 +1,34 @@
 #!/bin/bash
+# Running under WSL (Windows Subsystem for Linux)?
+if cat /proc/version | grep Microsoft; then
+    WSL_running=true
+else
+    WSL_running=false
+fi
 
 # TODO: remove the `-o Acquire::ForceIPv4=true` once Debian's mirrors work reliably over IPv6
 apt-get -o Acquire::ForceIPv4=true install -y sudo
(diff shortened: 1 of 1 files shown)


scottleibrand:
It'd be better if we could avoid duplicating everything.  One approach we've used elsewhere (like oref0-setup) is to define a die() function, and then have each line be `do the thing || die "couldn't do the thing"` rather than chaining a bunch of `this thing && that thing && the other thing`.  That would allow your if clause to just apply to the nodejs vs. nodejs-legacy decision.  Is that something you'd be comfortable refactoring as part of this change, or should we make it a separate thing?

sethgagnon:
Might be above my pay grade. 😀 I'll take a look though and let you know.

danamlewis:
Checking in on this - @sethgagnon did you have a chance to take a look?

sethgagnon:
Not sure the die function is in my wheelhouse, sorry. Maybe approve the code I submitted using the if else statements for now and then someone with stronger coding background can work on the die function?

scottleibrand:
Looks like this script is no longer used: the docs now reference the version in the oref0 repo instead.  There have also been some updates that might accomplish the same thing you're trying to do another way.  Can you try https://raw.githubusercontent.com/openaps/oref0/dev/bin/openaps-packages.sh and see if that works properly on WSL?

danamlewis:
Closing due to lack of activity; please re-open if you resume testing per the above suggestion. Thanks!

PR #1381, opened 2018-10-28
New phone info added to Edison bluetooth info

PedanticAvenger:
Added oneplus phone.


PedanticAvenger:
will re:pr this after a tweak.

PR #1387, opened 2018-11-24
Update monitoring-OpenAPS.md

audiefile:
Add note for reason offline webpage may not load and fix for it.

### DIFF
--- a/docs/docs/While You Wait For Gear/monitoring-OpenAPS.md
+++ b/docs/docs/While You Wait For Gear/monitoring-OpenAPS.md
@@ -583,7 +583,7 @@ The box around your current BG will be either green or red, depending on the las
 
 ![Offline webpage OpenAPS pill](../Images/offline_webpage_2.png)
 
-NOTE: If the webpage does not load, check your crontab. On master your crontab should contain the line `@reboot cd ~/src/oref0/www && export FLASK_APP=app.py && flask run -p 80 --host=0.0.0.0` You can check this by logging into your rig and typing `crontab -l`. If you need to edit your crontab the command is `crontab -e`.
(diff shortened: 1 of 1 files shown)


audiefile:
The offline webpage does not appear at all if that line is not in crontab. 

I'm on dev right now. Without that line I would get errors trying to access the site in browser, with the site will load. It will work with dev code but I'm not sure that I have the new style cron jobs.

scottleibrand:
Have you re-run oref0-setup since the 0.7.0-dev refactor?  If so, your crontab will just have a few lines, running oref0-cron-every-minute, oref0-cron-every-15min, oref0-cron-nightly, and oref0-cron-post-reboot. The latter has:

bin/oref0-cron-post-reboot.sh:    flask run -p 80 --host=0.0.0.0 | tee -a /var/log/openaps/flask.log

audiefile:
I did re-run the interactive setup last week after pulling what was then the latest dev branch. However, I still have a 'old' style crontab with a dozen or so lines.

scottleibrand:
Do you remember whether you deleted and replaced your crontab when you did oref0-setup?

audiefile:
I did not manually delete it and the script didn't ask about cron setup.

scottleibrand:
Sounds like oref0-setup didn't complete successfully then.  It should always prompt to replace crontab at the end.  I'd suggest re-running it.

audiefile:
I did. This time it did ask about cron but it's still the old style

scottleibrand:
If you don't delete the old crontab, it'll keep the old lines and add the new ones as well.  You'll want to say Yes to deleting the old crontab and let it recreate it.

audiefile:
I did remove the existing crontab.

scottleibrand:
If you're getting old-style cron entries re-added after removing the existing crontab, then you haven't successfully installed the latest oref0 dev before running oref0-setup.  The dev oref0-setup code is now this: https://github.com/openaps/oref0/blob/dev/bin/oref0-setup.sh#L1359-L1396

audiefile:
The new style crontab works

scottleibrand:
Cool.  Do you know what the problem was that prevented you from getting a new-style crontab earlier? 
Should we revise this PR to mention that as a better fix?  Or leave the docs as-is?

PR #1399, opened 2019-01-11
adding in 'sudo' where appropriate

iceman198:


scottleibrand:
The docs do specify that you should log in as root. Care to review them and see if there are any additional places where that should also be mentioned for clarity?

iceman198:
Thanks @scottleibrand .  If I have ideas, I'll run 'em by.

PR #1422, opened 2019-02-17
Changed pinout connection scheme

jansche:
Changed pinout connection scheme for RFM69HCW-only rig setups so it aligns with the radiofruit bonnet based rigs. As discussed in openaps/hardware-dev.


danamlewis:
@cluckj or @ecc1 can you +1 if this looks correct?

jansche:
ok, that's fair points. I'll re-do the PR with this in mind.

PR #1426, opened 2019-02-24
Fix the carriage returns in RFM69HCW

tzachi-dar:
Signed-off-by: Tzachi Dar <tzachi.dar@gmail.com>

Without this change, the lines on the md doc seems very hard to understand.

### DIFF
--- a/docs/docs/Gear Up/edison.md
+++ b/docs/docs/Gear Up/edison.md
@@ -100,10 +100,10 @@ Solder it to the board. It's the hole near the "o" from Radio. Make sure to not 
 
 This is your connection scheme for the RPi to RFM69HCW. Stick the RFM69HCW on a bread board, and connect:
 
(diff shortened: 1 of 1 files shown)


tzachi-dar:
After creating the PR, I have noticed that things look well before my change on github (for example on: https://github.com/tzachi-dar/docs/blob/master/docs/docs/Gear%20Up/edison.md) but not on 
https://openaps.readthedocs.io/en/latest/docs/Gear%20Up/edison.html.
However, most people look at openaps.readthedocs.io.
So, I guess that the simplest way around this, is to replace the text with a picture.

Does this sound reasonable?

danamlewis:
I would do both. (Essentially, add a picture below for clarity, but leave the text so it's easy to update it + generate a new picture in the future)

tzachi-dar:
Closing this, as https://github.com/openaps/docs/pull/1428 was opened.

PR #1430, opened 2019-02-28
Additional text for 640G CGM users as below

Teamhrach:
Some 640G users may (as I did) may choose the MDT option as their CGM during the set up. If this is done,then the algorithm will look to pull the BG data directly from the looping pump instead of NS (which gets its data from the 640G)   

'Note:  For Medtronic 640G (CGM) users, it is recommended that you enter 'xdrip' - otherwise the BG values may not be read from your Nightscout. (The reason being, the 'MDT' option applies only for the enlite sensor attached to the actual pump you're looping with)'


Teamhrach:
Feeling proud  [/var/folders/j0/wwrhp4fd0kjdq16z_bx4rt3w0000gn/T/com.microsoft.Outlook/Content.MSO/8C14304E.tmp]

PR #1440, opened 2019-03-13
Hardware / supplies safety notice

sulkaharo:
This could also be in the hardware section, but it might be better to be really, really up front about this?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -15,6 +15,12 @@ Note: *We do not recommend using a PDF version of this guide. The docs are updat
    
    This community of contributors believes in "paying it forward," and individuals who are implementing these tools are asked to contribute by asking questions, helping improve documentation, and contributing in other ways. Have questions? Hop into `Gitter <https://gitter.im/nightscout/intend-to-bolus>`_ and ask anytime!
 
+.. DANGER:: 
+   **IMPORTANT SAFETY NOTICE**
+
+   The foundation of OpenAPS safety features discussed in this documentation are built on the safety features of the hardware used to build your system. It is critically important that you only use a tested, fully functioning FDA or CE approved insulin pump and CGM for closing an automated insulin dosing loop. Hardware or software modifications to these components can cause unexpected insulin dosing, causing significant risk to the user. If you find or get offered broken, modified or self-made insulin pumps or CGM receivers, *do not use* these for creating an OpenAPS system.
+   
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -15,6 +15,12 @@ Note: *We do not recommend using a PDF version of this guide. The docs are updat
    
    This community of contributors believes in "paying it forward," and individuals who are implementing these tools are asked to contribute by asking questions, helping improve documentation, and contributing in other ways. Have questions? Hop into `Gitter <https://gitter.im/nightscout/intend-to-bolus>`_ and ask anytime!
 
+.. DANGER:: 
+   **IMPORTANT SAFETY NOTICE**
+
+   The foundation of OpenAPS safety features discussed in this documentation are built on the safety features of the hardware used to build your system. It is critically important that you only use a tested, fully functioning FDA or CE approved insulin pump and CGM for closing an automated insulin dosing loop. Hardware or software modifications to these components can cause unexpected insulin dosing, causing significant risk to the user. If you find or get offered broken, modified or self-made insulin pumps or CGM receivers, *do not use* these for creating an OpenAPS system.
```
scottleibrand: Does this also apply to DIY CGM receiver software?

danamlewis: I'm wondering about scratching the last sentence to make this more broadly relevant, since it's a reword and repeat of the above sentences. @sulkaharo thoughts?

scottleibrand: My specific concern is that "software modifications to these components can cause unexpected insulin dosing, causing significant risk to the user" is equally true of DIY CGM receiver software that implements non-FDA-approved calibration algorithms.  If we're going to take a principled stance against "modified ... CGM receivers" vs. "FDA or CE approved ... CGM", it seems that conflicts with our support of xDripAPS and Spike.

scottleibrand: Oh, and as written this would also be a very strong warning against using rebatteried Dexcom transmitters, which AFAICT are perfectly safe.

sulkaharo: Wonder how to parse this... 1) I guess this should be generic in a manner where if someone invents some new stupid idea, it's covered, but also 2) if we know something is safe(ish?), it's not described as dangerous, but also 3) implying any hacks are "safe" is somewhat wrong.  We actually _don't_ know rebatteried CGMs are safe - I just changed between transmitters after it started to seem to me the one that'd been rebatteried several times was producing more noise than before and indeed the new transmitter seems to be much less noisy.

I guess something that'd maybe clarify this is being specific about firmware mods?

I would like to mention non-approved pumps specifically due to the crazy fellow selling his self-made pumps.

Issue #1447, opened 2019-03-27
Debian Jessie has lost upstream support


cluckj:
The package mirrors and apt repositories for debian jessie have been deprecated: https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html

We should probably change the docs to recommend jubilinux 0.3 (stretch), and recommend reflashing for folks who are using jubilinux 0.2 (jessie).

(I'm creating for discussion of the issue, and to track all the changes we need to make to the docs)

Related: https://github.com/openaps/oref0/issues/1236

danamlewis:
@cluckj

PR #1448, opened 2019-03-28
Migrate to Jubilinux 0.3.0 (stretch)

cluckj:
- Updates Edison install docs to recommend Jubilinux v0.3.0 (stretch) install, rather than v0.2.0 (jessie), which has its upstream support deprecated (#1447 / https://github.com/openaps/oref0/issues/1236). 

- Switches the Edison bootstrap script to point to the `dev` branch, as `master` does not currently work on v0.3.0 (this should be reverted if there is a workaround, or once https://github.com/openaps/oref0/pull/1036 is merged)

### DIFF
--- a/docs/docs/Build Your Rig/edison-install.md
+++ b/docs/docs/Build Your Rig/edison-install.md
@@ -48,8 +48,8 @@ ifdown wlan0; ifup wlan0
 sleep 10
 echo -ne "\nWifi SSID: "; iwgetid -r
 sleep 5
-curl https://raw.githubusercontent.com/openaps/oref0/master/bin/openaps-install.sh > /tmp/openaps-install.sh
-bash /tmp/openaps-install.sh
+curl https://raw.githubusercontent.com/openaps/oref0/dev/bin/openaps-install.sh > /tmp/openaps-install.sh
+bash /tmp/openaps-install.sh dev
 )
 ```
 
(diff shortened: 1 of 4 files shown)


scottleibrand:
Does this effectively force everyone onto dev?  Would we get the same outcome merging dev to master?

cluckj:
It would only put folks on the Edison onto dev, but...yes.

danamlewis:
Do we still want this?

scottleibrand:
I think we still want the parts that tell people to use jubilinux 0.3.0 instead of 0.2.0.  Not sure about the edison-install.md changes to use dev instead of master.  Perhaps those aren't needed once we merge https://github.com/openaps/oref0/pull/1241 ?

cluckj:
Yep, I think the edison-install.md change is unnecessary after merging https://github.com/openaps/oref0/pull/1241 The "does not work anymore" language might need to be changed (though that will encourage folks to reflash with 0.3.0...)

daviddrozh:
Hi, I just happened to somehow stumble on this thread. I've been trying to flash my edison with 0.3.0 for the past 5 hours, and keep hitting a brick wall with the bootstrap script. I've narrowed it down to not being able to install nodejs-legacy and npm. 
I've tried working around it by installing through nvm, and skipping to step 4 of the install process, but no luck with either routes. Any thoughts on how i can fix this?

scottleibrand:
For master, try https://gist.github.com/scottleibrand/9372092ac914c7edbe23fa4eb94b83d8

For dev, try:
```
curl https://raw.githubusercontent.com/openaps/oref0/dev/bin/openaps-install.sh > /tmp/openaps-install.sh
bash /tmp/openaps-install.sh dev
```

PR #1454, opened 2019-04-16
Pass `dev` to openaps-install.sh

cluckj:
If we want to install `dev`, we should tell openaps-install.sh to pull and install the `dev` branch.

### DIFF
--- a/docs/docs/Build Your Rig/pi-install.md
+++ b/docs/docs/Build Your Rig/pi-install.md
@@ -60,7 +60,7 @@ Once you're logged in, run the following commands to start the OpenAPS install p
 
 ```
 sudo bash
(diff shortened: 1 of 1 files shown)


scottleibrand:
Do we actually want to install `dev` here?  Aren't these the commands run by default by users installing master?  It looks like those have always pointed to the `dev` version of the install script (probably because we never bothered to change it to `master`, and never noticed because they're similar enough that it hasn't mattered), but AFAICT they've always installed `master`.

cluckj:
Yeah I think you're right; the issue this was trying to solve is turning out to be something else.

cluckj:
iirc it defaults to installing `master` if no branch is specified, I'll close this out.

PR #1455, opened 2019-04-16
Tzachi dar adafruit bonnet

tzachi-dar:
This changes the wiring of RFM69HCW from the "walrus" format to "bonnet" format, which is the one they should be using.


ecc1:
1. I see two copies of the wiring table in edison.md
2. The column heading for pin 26  "CE0_N" should be "CE1_N" (chip enable 1, negated)

A larger question: shouldn't the beginning of this section mention the Adafruit RFM69 bonnet as a (preferable?) option, since no wiring is required and it should "just work"?

tzachi-dar:
1) In github there are two pictures. On the docs page, the first one is not clear, so I also add a table of it.
See https://openaps.readthedocs.io/en/latest/docs/Gear%20Up/edison.html#soldering for how most people will see it.

2) Update pin name.

As for the larger question...
I was not aware of this option for example... Spent a lot of time soldering things.
Indeed if it is as simple as that, then I can add the text if people will help.

Is it as simple as:

RFM69HCW
The simplest way to build a RPII based rig is using Adafruit Radio Bonnet 915 MHZ. You can get it here for example (or anywhere you want). This option does not need soldering, has a screen and buttons and  is ready to use (we can add a few words about antenna and so on).

If you want a somewhat chipper/smaller option then -- and here we continue with soldering.

PR #1470, opened 2019-05-27
Adding description of new Adafruti board with RFM69HCW

docx1:
-added description with the product link
-reference to the 3D case on Thingiverse
-suggested installation cautions


danamlewis:
covered via another commit; adding 3d case link separately. thanks!

PR #1475, opened 2019-06-09
Update bluetooth instructions for Debian Stretch

cluckj:
Since we are using Debian's bluetooth package on Stretch, we don't need to manually kill/start bluetoothd.


scottleibrand:
Should we hide the jubilinux-0.2.0-only content behind a > arrow that you have to expand to see it?

PR #1479, opened 2019-06-13
Update troubleshooting for "Monitor/mmtune.json is empty or does not …

davidvjr:
…exist" error

I had red and green LEDs on both of my rigs and the "Monitor/mmtune.json is empty or does not exist" error in my loop log. Both rigs were  running dev for several weeks before any problems. I used the above method to fix them both (loops started working again on master branch, have not switched to dev yet). This procedure has obviously not been vetted thoroughly, so it may be of limited use for the docs, but I figured I would document it and it might be a starting point if this is a commonly occurring problem. 

This is my first PR so I tried to integrate it into the existing troubleshooting section to the best of my ability. I also linked to other sections of the docs. I'm not sure how close I came to formatting this correctly, but it's better than inaction I suppose!

### DIFF
--- a/docs/docs/Resources/troubleshooting.md
+++ b/docs/docs/Resources/troubleshooting.md
@@ -250,6 +250,7 @@ cd ~/src/ccprog
 ./ccprog -p 19,7,36 write spi1_alt2_EDISON_EXPLORER_US_STDLOC.hex
 ```
 
+
 #### Using a Raspberry Pi + Explorer HAT:
 ```
 wget https://github.com/EnhancedRadioDevices/subg_rfspy/releases/download/v0.8-explorer/spi1_alt2_EDISON_EXPLORER_US_STDLOC.hex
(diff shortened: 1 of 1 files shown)


davidkeddydb:
Just reflashed one of my rigs using this method.  Not successful.  Still have blinking radio lights after reflash and rig is looping.

f87m19:
Thank you for this. I have run into this problem more than once. Pointed in the right direction by this PR, I have now learned that when you get this error in 'dev' (on Edison), a method to fix the problem is:
- Flash the Edison
- Install 'master'
- Use instructions in this PR to reflash radio
- Install 'dev' (given that one wants to continue using dev)

I did not succeed in fixing this directly working in dev.

davidkeddydb:
I have done this with rig, installed master and flashed chip. My radio lights still flash however before and after I updated to dev.  I am looping, just blinking at same time.  Have not tried yet on second rig which is also doing this.  I did see the chip say erased and flashed however fwiw.

scottleibrand:
The LEDs are *supposed* to flash while you’re looping, every time a new command is run.

davidkeddydb:
@scottleibrand I guess I didn't remember them flashing repeatedly for say 5-6 blink rotations every 60-90 seconds but maybe they did and I forgot? Sorry for my confusion.

scottleibrand:
The normal LED behavior is to flash once every 1 seconds (as it prints each `.`) during wait-for-silence, and then once for each other radio command.  So if you watch the pump-loop.log and the LEDs at the same time, you should see them flash every time something new is written to the log.

PR #1495, opened 2019-09-15
iOS Shortcut to Nightscout integration instructions

endafarrell:
Since summer 2018 the iOS app "Shortcuts" (previously Workflow) dropped support for IFTTT. We can get same effect with these instructions.


endafarrell:
+1

danamlewis:
fantastic. thank you @endafarrell!

endafarrell:
Thanks all. For an awful lot :-)

PR #1503, opened 2019-10-24
Fix the use of mmtune and the use of killall.

tzachi-dar:


tzachi-dar:
OK, changed all instances of killall -g to killall-g

Please review.

PR #1514, opened 2019-11-15
edison.md: Convert numbers to actual numbered list

StephenBrown2:


danamlewis:
We have done this in the past, but the numbering breaks in display on the RTD rendered view, so we'd prefer to leave with the 'weird' formatting for that reason. Thanks!

PR #1517, opened 2019-11-18
Mac flashing instructions update

straykatz:
1) Added reminders to connect battery before flashing to reduce risk of power glitch problems during flashing and UNIX upgrade. 
2) Updated instructions concerning screws and spacers, batteries, use of USB-C cables (no hubs, no adapters; when in doubt, try new cables)
3) Deleted a concern about using new MacBooks with USB-C ports, as issue is covered in the USB-cable information.


straykatz:
It's probably two commits because it took me a while to figure out how to edit a commit. :-)

PR #1520, opened 2019-11-25
there is no step 4

dougboss:
"If you are using Jubilinux 0.3.0 (Debian Stretch) or the Raspberry Pi, please skip to #4. If you are using Jubilinux 0.2.0 (Debian Jessie), you will need to manually initialize bluetooth. (click here to expand instructions)"

There is no step 4 that  I can find ... 
What are the steps for  Jubilinux 0.3.0 ?


danamlewis:
@dougboss resolved via https://github.com/openaps/docs/pull/1521, the docs are now updated (the formatting dropped). PS - in future you can open an issue in https://github.com/openaps/docs/issues if you spot issues with the docs but don't have a proposed fix. Thanks for flagging it!

PR #1521, opened 2019-11-25
Formatting fix

cluckj:
Fixes formatting for RTD, as noted in #1520. Line 118 (`<summary>` block) doesn't work yet...


docs/docs/Customize-Iterate/bluetooth-tethering-edison.md
```diff
@@ -4,13 +4,13 @@ Your cell phone can act as a mobile "hotspot" to allow your rig to access the in
...
 
-2. Hotspot, when activated, uses your cell phone's data.  Know what your cell phone plan data limits are and consider if you want to change/update based on your frequency of hotspot use.  You can get an estimate of cell data use by resetting your cell data use, at the beginning of the day, within your phone.
+`2.` Hotspot, when activated, uses your cell phone's data.  Know what your cell phone plan data limits are and consider if you want to change/update based on your frequency of hotspot use.  You can get an estimate of cell data use by resetting your cell data use, at the beginning of the day, within your phone.
 
-3. A device (like your rig) can be connected to your phone's hotspot in one of three ways:
+`3.` A device (like your rig) can be connected to your phone's hotspot in one of three ways:
```
jakobsandberg: Using inline-code formatters here means this line is no longer interpreted as an ordered list item, which has unintended consequences:

The following lines '**BT tether** ...' gets treated as a code block instead of a paragraph. Thus, you need to scroll horizontally to read the entire thing.

docs/docs/Customize-Iterate/bluetooth-tethering-edison.md
```diff
@@ -100,7 +100,7 @@ root@edisonhost:~# bluetoothd --version
 
 ### Bluetooth setup
 
-1) First, check that your wpa_supplicant.conf file doesn't contain any content that will interfere with oref0-online.
+`1.` First, check that your wpa_supplicant.conf file doesn't contain any content that will interfere with oref0-online.
```
jakobsandberg: Formatting this ordered list item as code also has unintended side effects on both lines (a) and (b) below. Notice that:
(a) The whole line is being treated as a code block, even we have wrapped inline-code formatters around the text. Thus, when readers copy-and-paste this in their terminals they will get errors because it will include the inline code formatters.
(b) This line is also being treaded as a code block, and thus isn't being parsed as an image. Thus, we no longer display the png image on the doc.
(b) sdfsdf

danamlewis: This was caused by extra spacing, rather than the formatting of the list items directly. Resolved.

danamlewis:
> By wrapping our ordered list items in inline-code formatted (backticks ``) our markdown is interpreted very differently, which breaks some existing functionality.
> 
> What is the reason for treating these ordered list items as inline-code?

The lists often break in RTD with extra spacing, so using the backtick formatting seems to work better for maintaining numbered lists without causing confusion in the future. Thanks for flagging the other issues, @jakobsandberg , those are also now resolved (removing the extra spacing, unrelated issue).

PR #1550, opened 2020-06-19
Fixed DIA source

loztblaz:
Per @danamlewis on gitter on 06-04-2019, DIA is pulled from the pump if it is over 5h, otherwise it defaults to 5h. Haven't edited docs before, so please suggest changes if the wording could be better or if I'm doing anything wrong with this PR.


danamlewis:
Thanks for the PRs, @loztblaz !

PR #1560, opened 2020-11-12
Update index.rst

old-square-eyes:
Include flashing troubleshooting page


scottleibrand (approved):

old-square-eyes:
Thanks @scottleibrand. I see this is approved. How do I merge?

danamlewis:
For future navigation changes: there are two places to change navigation. One is in the sub-section per @scottleibrand initial suggestion; there is also https://github.com/openaps/docs/blob/master/docs/index.rst which I changed in your branch to include in this PR as well. That gets both the home page nav and also the left menu bar nav updated.

PR #1579, opened 2022-02-19
change position of switch argument

KelvinKramp:
On my computer and also on ubuntu ec2 instance the line: "oref0-upload-profile --switch ./myopenaps/autotune/profile.json $NS_SITE $API_SECRET " only works if you change the orderof the arguments to "oref0-upload-profile ./myopenaps/autotune/profile.json $NS_SITE $API_SECRET --switch=true"


scottleibrand (approved):

scottleibrand:
The usage string is `$0 profile.json NSURL api-secret [--preview] [--switch]`, so this looks correct. The `--switch` argument doesn't appear to be used anywhere in oref0, and seems to just be for manual use.

PR #1584, opened 2022-09-21
Update autotune.md

juehv:
Add link to autotune on aws to the documentaton


danamlewis:
@juehv I notice the comment in your repo that you're not planning to keep this updated - checking to see if that's true? if so, might tweak where this is placed in the docs as a result.

PR #1589, opened 2023-02-14
Second try at correcting the table

inventor96:
I didn't get to see what it looked like before the backticks were added (https://github.com/openaps/docs/commit/1dba918aa63bb0c616c9b61c3af933a0ac2eb651), so I don't know if it was enough to get the Read the Docs page formatted correctly, but obviously there's something else wrong since the PDF stuff broke. This is my second attempt. I realized that there were some inconsistencies in the use of spacing characters, and an extra column on one of the rows. Hopefully this attempt will work better. :crossed_fingers:

### DIFF
--- a/docs/docs/Gear Up/edison.md
+++ b/docs/docs/Gear Up/edison.md
@@ -106,13 +106,11 @@ Solder it to the board. It's the hole near the "o" from Radio. Make sure to not 
 
 This is your connection scheme for the RPi to RFM69HCW. Stick the RFM69HCW on a bread board, and connect:
 
-```
 | Board | Connect | Connect | Connect | Connect | Connect | Connect | Connect | Connect |
-| ------|------|------|------|------|------|------|------|------ |
-| RPi	| 3.3V	| GND	| MOSI | MISO | SCLK	| | CE1_N	||  |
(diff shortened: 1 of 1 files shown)


trevettallen:
Wow 😳😀😀😀

trevettallen:
You're doing REALLY cool stuff ☺️☺️☺️☺️

I want to understand better

danamlewis:
@inventor96 sorry, the backticks were added by me! after I merged the PR, something in the RTD build broke, so I was trying to see if containing the table fixed it. I fixed the broken build issue separately but hadn't gone back to remove the backticks yet. 

I can preview PRs in RTD now - here's a screenshot of how this version will render. It's still not turning into a table yet, but I'll go ahead and merge this regardless as it undoes the backticks and fixes the things you mentioned.

Screen Shot 2023-02-14 at 9 50 09 AM

inventor96:
@danamlewis Yeah, I edited my comment after seeing finding the commit. 👍

It's a bummer the table still doesn't layout properly. It looks like we use HTML tables elsewhere in the docs, rather than Markdown tables, so I guess I'll create yet another PR and use that method instead.

